PR #1, opened 2022-03-23
Feature/fielded optional allow null fill with default values

aliothe:
FieldedTypeBuffer:

Allow null values that are written to be written using a set of predefined default values.

Note that this breaks the symmetry with the c++ side regarding how it works when writing fielded.

We allow this since it seems to be common in old code that people want to be able to write null values ( even though you really can not since we can not transport it).


aliothe:
@TobiasLeo @cklellie

PR #8, opened 2022-04-05
follow the @Timeout contract

aliothe:


aliothe:
The contract:
"Methods annotated @Timeout in the enterprise bean class must return void and optionally take a javax.ejb.Timer object as the only parameter. They may not throw application exceptions."

TobiasLeo (approved):

PR #16, opened 2022-05-01
Feature/casual caller better queue handling

### LINKED ISSUE #12: casual-caller better queue handling
Casual queue needs to take care about selecting queues from multiple pools if multiple casual instances handle the a queue by the same name. This should normally not happen and could be seen as a config error, but if it does casual-caller needs to stick to just one instance of that queue for all queue operations

Also extend the queue api to use EnqueueReturn and DeqeueuReturn wrapper types for queue operations to enable error state reporting similar to the service api (we mainly want to be able to report TPENOENT right now for queues that do not exist in any casual connection known to casual-caller

TobiasLeo:
- Updated queue api with breaking changes (both for caller and jci because of shared api). Now using wrapper for enqueue/dequeue to handle metadata like error state (only that for now, but could possibly accomodate for others in the future)
- Can lookup queues even if casual pools are unavailable, will return with errorstate TPENOENT if the queue could not be found (it could not exist in any pool, or the pool could be down
- Connection factories for queues are stickied to a specific connection pool per queue name. Normally there shouldn't be multiple sources for one queue, but if there are then the first encountered pool with the queue will be selected and stickied. The order seems to be pretty stable, but different application servers yield different orders. Never assume that the selected pools for a queue with multiple sources will be stable over different runs of the application server or if the queue cache is cleared

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/api/CasualQueueApi.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/CasualQueueApi.java
@@ -6,12 +6,11 @@
 
 package se.laz.casual.api;
 
+import se.laz.casual.api.queue.DequeueReturn;
+import se.laz.casual.api.queue.EnqueueReturn;
 import se.laz.casual.api.queue.MessageSelector;
 import se.laz.casual.api.queue.QueueInfo;
 import se.laz.casual.api.queue.QueueMessage;
-
-import java.util.List;
-import java.util.UUID;
 
 /**
  * Interface to operate on a casual queue
@@ -23,18 +22,18 @@ public interface CasualQueueApi
      * @param qinfo the queue info
      * @param msg the queue message
      * @throws se.laz.casual.network.connection.CasualConnectionException
-     * @return the id of the enqueued message
+     * @return a wrapper containing the uuid of the enqueued message and an ErrorState
      */
-    UUID enqueue(QueueInfo qinfo, QueueMessage msg);
+    EnqueueReturn enqueue(QueueInfo qinfo, QueueMessage msg);
 
     /**
      * Dequeue message(s)
      * @param qinfo the queue info
      * @param selector the queue message
      * @throws se.laz.casual.network.connection.CasualConnectionException
-     * @return a list of matching messages
+     * @return a wrapper containing a list of matching messages and an ErrorState
      */
-    List<QueueMessage> dequeue(QueueInfo qinfo, MessageSelector selector);
+    DequeueReturn dequeue(QueueInfo qinfo, MessageSelector selector);
 
     /**
      * Check if queue exists
--- a/casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2022, The casual project. All rights reserved.
+ *
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+
+package se.laz.casual.api.queue;
+
+import se.laz.casual.api.flags.ErrorState;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class DequeueReturn
+{
+    private final QueueMessage queueMessage;
+    private final ErrorState errorState;
+
+    private DequeueReturn(QueueMessage queueMessage, ErrorState errorState)
+    {
+        Objects.requireNonNull(errorState, "errorState can't be null");
+        this.queueMessage = queueMessage;
+        this.errorState = errorState;
+    }
+
+    public Optional<QueueMessage> getQueueMessage()
+    {
+        return Optional.ofNullable(queueMessage);
+    }
+
+    public ErrorState getErrorState()
+    {
+        return errorState;
+    }
+
+    public static Builder createBuilder()
+    {
+        return new Builder();
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o)
+        {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass())
+        {
+            return false;
+        }
+        DequeueReturn dequeueReturn = (DequeueReturn) o;
+        return Objects.equals(queueMessage, dequeueReturn.queueMessage) && errorState.equals(dequeueReturn.getErrorState());
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return queueMessage.hashCode() + Integer.hashCode(errorState.getValue());
+    }
+
+    @Override
+    public String toString()
+    {
+        return "DequeueReturn{" + "queueMessage=" + queueMessage +
+                ", errorState=" + errorState.name() + '(' + errorState.getValue() + ')' +
+                "}";
+    }
+
+    public static final class Builder
+    {
+        private QueueMessage queueMessage;
+        private ErrorState errorState;
+
+        public Builder withQueueMessage(QueueMessage queueMessage)
+        {
+            this.queueMessage = queueMessage;
+            return this;
+        }
+
+        public Builder withErrorState(ErrorState errorState)
+        {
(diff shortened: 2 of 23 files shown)


casual/casual-api/src/main/java/se/laz/casual/api/CasualQueueApi.java
```diff
@@ -25,7 +27,7 @@
      * @throws se.laz.casual.network.connection.CasualConnectionException
      * @return the id of the enqueued message
```
aliothe: Does not return the id of the enqueued message anymore, rather a EnqueueReturn

casual/casual-api/src/main/java/se/laz/casual/api/CasualQueueApi.java
```diff
@@ -34,7 +36,7 @@
      * @throws se.laz.casual.network.connection.CasualConnectionException
      * @return a list of matching messages
```
aliothe: Does not return a list of messages anymore

casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public List<QueueMessage> getQueueMessages()
+    {
+        if (ErrorState.OK.equals(errorState))
+        {
+            return queueMessages;
+        }
+        else
+        {
+            throw new CasualRuntimeException("Attempt to get queuemessages from queuereturn, but state was not OK");
```
aliothe: Add the actual error in the message

aliothe: Not sure we should go with a getter that can throw.
From a usage point of view, I'd not except that that can happen - an empty list, fine and if so then it is up to me, the user, to check the error state.

TobiasLeo: I did that to possibly prevent hiding errors, but I should probably make it return an empty list and leave it to the user to check ErrorState or not. In a way returning an empty list would be more similar to EnqueueReturn that returns Optional.empty(), but at the same time an empty list does not convey that something is amiss. I'll probably go with empty list and add some javadoc to clarify usage for any curious user

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryLookupService.java
```diff
@@ -24,20 +25,20 @@ public class ConnectionFactoryLookupService implements ConnectionFactoryLookup
...
-            return cachedEntries;
+            return cachedEntry;
         }
         List<ConnectionFactoryEntry> newEntries = lookup.find(qinfo, connectionFactoryProvider.get());
         if (!newEntries.isEmpty())
         {
             cache.store(qinfo, newEntries);
         }
-        return newEntries;
+        return cache.getSingle(qinfo); // May be something or empty depending on if
```
aliothe, line 41: depending on if ?

casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
cklellie: But if you retrieve the messages and there are none, from a user perspective not having it raise the error is a bit worrying to me. I prefer this initial implementation. 
Question - can you have messages and an error state? If yes, then move the error checking out - if not, then I would say leave it in.

aliothe: I think that it is not exceptional to ask for messages from a queue, there are no messages, you get an empty list back.

aliothe: Then again, that will also be covered by the ErrorState being OK

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryLookup.java
```diff
@@ -30,7 +31,7 @@
      * @param qinfo
      * @return a List of 0-n CacheEntries
```
cklellie: This is a single optional now, was a list - is that correct? 
If so update the javadocs to reflect.

casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
aliothe: Still feels weird with a getter that throws but at least we still cover the case where there are no messages, that was the one I was considering

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/jmx/CasualCallerControl.java
```diff
@@ -6,57 +6,107 @@
...
-    public CasualCallerControl(Cache cache)
+    public CasualCallerControl(Cache cache, ConnectionFactoryEntryStore connectionFactoryEntryStore)
     {
         this.cache = cache;
+        this.connectionFactoryEntryStore = connectionFactoryEntryStore;
     }
 
     @Override
-    public void purgeServiceCache()
+    public List<String> poolsValid()
```
cklellie: You have changed the ordering of the words for some of the methods which for me makes them not nice to read.
action / adjective then noun is much nicer to read:
validPools rather than poolsValid
invalidPools rather than poolsInvalid
purgeServiceCache rather than serviceCachePurge
etc

versions.gradle
```diff
@@ -7,10 +7,10 @@
 
 // casual-java group and version
 group = 'se.laz.casual'
-version = '1.0.82-beta'
+version = '1.0.0'
```
cklellie, line 10: Accidental reset?

casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
TobiasLeo: If the call to dequeue resulted in an error (currently that would be TPENOENT when the queue isn't accessible in any connection pool) there should never be any data to retrieve from the return wrapper. Throwing absolutely makes sense here. After some thought I'll probably keep the throwing, but add more info to the exception message. I'll also add some javadoc to the method to clarify expected behavior to users

Compared to the old behavior both enqueue and dequeue will require extra work for the end user. Based on my tests enqueue previously returned an all zero UUID if the queue did not exist and dequeue would return an empty list. While it's more work for the end user I do think it's better in the long run that errors are exposed and states with errors need to be handled explicitly

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryLookup.java
TobiasLeo: Yes, this is correct. I'll update the javadocs. Normally there should never be more than one configured source for a queue in a system, but if there is we will always sticky the first one we encounter and continue to use that one (until the server running casual-caller is restarted/the queue cache is cleared). That is apparently how casual itself handles multiple queues of the same name more or less

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/jmx/CasualCallerControl.java
TobiasLeo: For this one I was on the fence but finally elected to do this way to have the methods sort by category. After using these myself i have to agree these are not fun to read at all. I'll go change it back to names that make sense to read

versions.gradle
TobiasLeo: This change is intentional. Since I'm making breaking api changes to CasualQueueApi (which propagates to CasualCallerApi) the version should reasonably change on the major. After a previous discussion with @AdjacentDoor we concluded that it could make sense to use the occasion instead to drop the beta and just call this 1.0.0, while casual-caller was bumped to 2.0.0 since it already had a "final" version 1.*

aliothe (approved):

PR #21, opened 2022-05-06
Feature/casual caller needs to handle connection reestablished better

### LINKED ISSUE #20: casual-caller needs to handle connection reestablished better
The issue is that if a connection has been used for n number of service calls, this is then cached and no further service discovery is issued towards casual once a connection is reestablished.

Maybe it could be handled in such a way that when a connection is reestablished, goes from invalid to valid, we'll issue one domain discovery request towards that casual resource containing the currently known world with regards to queues and services.
The result is then cached for this connection.

aliothe:
* When a previously known connection goes away and comes back again, rediscovery of all known services/queues ( for all connections) needs to be issued and the caches repopulated with the result of that discovery.

### DIFF
--- a/casual/casual-caller/src/main/java/se/laz/casual/connection/caller/Cache.java
+++ b/casual/casual-caller/src/main/java/se/laz/casual/connection/caller/Cache.java
@@ -32,7 +32,7 @@ public class Cache
 
     public List<ConnectionFactoryEntry> get(QueueInfo qinfo)
     {
-        return Optional.ofNullable(queueCache.getAll(qinfo)).orElseGet(() -> (Collections.emptyList()));
+        return Optional.ofNullable(queueCache.getAll(qinfo)).orElseGet(Collections::emptyList);
     }
 
     public Optional<ConnectionFactoryEntry> getSingle(QueueInfo qinfo)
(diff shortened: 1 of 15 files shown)


casual/casual-caller/src/main/java/se/laz/casual/connection/caller/Cache.java
```diff
@@ -27,7 +32,7 @@ public ConnectionFactoriesByPriority get(String serviceName)
 
     public List<ConnectionFactoryEntry> get(QueueInfo qinfo)
     {
-        return queueCache.getAll(qinfo);
+        return Optional.ofNullable(queueCache.getAll(qinfo)).orElseGet(() -> (Collections.emptyList()));
```
TobiasLeo: Could use a method reference instead of lambda

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/Cache.java
```diff
@@ -60,6 +65,33 @@ public void purgeQueues()
...
+    {
+        Map<CacheType, List<String>> entries = new EnumMap<>(CacheType.class);
+        entries.put(CacheType.SERVICE, getServices());
+        entries.put(CacheType.QUEUE, getServices());
+        return entries;
+    }
+    public void repopulate(DiscoveryReturn discoveryReturn, ConnectionFactoryEntry connectionFactoryEntry)
+    {
+        discoveryReturn.getServiceDetails().stream()
+                       .forEach(serviceDetails -> {
```
TobiasLeo: Any specific reason to use Stream::forEach and not Iterable::forEach on the list directly?

TobiasLeo (commented):
Just some minor points, but otherwise the solution looks sound. Most of it should be reusable in the future when casual starts to send out topology change events that we will want rediscoveries for as well

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/Cache.java
aliothe: fixed

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/Cache.java
aliothe: nope, changed to Iterable::forEach

TobiasLeo (approved):

PR #25, opened 2022-05-15
Handling case in casual-caller where exception is thrown, but should be TPENOENT

### LINKED ISSUE #24: Casual caller throws exception when it should send TPENOENT reply
If casual caller has a pool with a known, cached service and then all pools that handle that service are for some reason offline, then an exception is thrown when an attempt is made to call that service. The correct behavior should be to return a TPENOENT reply

TobiasLeo:
When a service is known in the cache, but no valid pool handles it an exception is thrown incorrectly. Instead a TPENOENT reply should be returned. This should now mirror the behavior where a that is completely unknown is unreachable

(and some minor fixes discovered while testing this ...)
Making sure services rediscovered on a validation rediscovery are tagged as checked as well as found
Fixing issue where attempted discoveries containing only a list of checked/resolved pools never registered them as checked. This is most likely to happen if not all pools are discovered together


aliothe (approved):
Looking good

PR #27, opened 2022-06-03
no optional parameter

### LINKED ISSUE #28: Remove use of optional as method params
Noticed that the Conversation interface has an Optional in the method parameters.
Optional should only be used as return types.
https://rules.sonarsource.com/java/RSPEC-3553
Preferably use overloads, builders, varargs or other strategies.

Mathias created a PR when we found the first instance, though we quickly found another. in CasualConversationAPI.
As these changes in interfaces will result in breaking changes therefore a major version bump, should check rest of code base and fix in one go.
There are currently 245 instances of Optional uses in the code base which we need to go through and assess / fix.

Additional question:
Why was not picked up in sonar scanning prior to commiting?

aliothe:
As this breaks the API, we bump to 2.0.0

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/FieldedTypeBuffer.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/FieldedTypeBuffer.java
@@ -581,7 +581,7 @@ public final class FieldedTypeBuffer implements CasualBuffer
 
     /**
      * Creates a {@code Supplier<CasualFieldedLookupException>}
-     * with the message that the {@code name} is missing at {@code Optional<Integer>} index
+     * with the message that the {@code name} is missing at {@code Integer} index
      * Index is 0 if not supplied
      * @param name the name
      * @param index the index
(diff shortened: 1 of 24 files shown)


aliothe:
@cklellie double check please that there are no more Optional as params, not even in private or package private contexts where we can guarantee consumption - it's all null everywhere hopefully.

casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/FieldedTypeBuffer.java
```diff
@@ -572,6 +573,12 @@ public List<byte[]> getBytes()
...
 
+
+    public static Supplier<CasualFieldedLookupException> createNameMissingException(String name)
+    {
+        return createNameMissingException(name, null);
+    }
+
     /**
      * Creates a {@code Supplier<CasualFieldedLookupException>}
      * with the message that the {@code name} is missing at {@code Optional<Integer>} index
```
cklellie: Fix comment. Also one on row 601, but i can't comment that low.

cklellie (approved):

cklellie:
> @cklellie double check please that there are no more Optional as params, not even in private or package private contexts where we can guarantee consumption - it's all null everywhere hopefully.

183 occurances, none as in params. Few places where usage could be cleaner imo. Have messaged you on slack. Can think about that for later.

Issue #31, opened 2022-07-03
For each connection in the same pool, they can all be connected to different casual servers


aliothe:
Thus casual caller need to be able to cache on the domain connect reply domain id.
It also needs to be able to add itself as observer for disconnect events regardless of reason ( appserver destroy/close or network error).

casual-jca needs to expose the domain id from the domain connect reply msg as well as the ability to add yourself as an observer for the physical connection.

PR #32, opened 2022-07-05
expose casual domain id and connection gone

aliothe:
See https://github.com/casualcore/casual-java/issues/31

This is the first part to solve that issue.
The second part is to change casual-caller to make use of this.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
@@ -24,8 +24,19 @@ public interface CasualConnection extends CasualServiceApi, CasualQueueApi, Casu
     @Override
     void close();
 
+    /**
+     * Get the current domain id for this connection
+     * @return the current domain id for this connection
+     */
     DomainId getDomainID();
 
+    /**
+     * If a connection is gone, for whatever reason - it is just not usable anymore.
+     *
+     * When this happens, {@link ConnectionObserver#connectionGone(DomainId)} is called with the same domain id
+     * that can be retrieved via {@link CasualConnection#getDomainID()}
+     * @param observer - the observer that will be notified
+     */
     void addConnectionObserver(ConnectionObserver observer);
 
 }
--- a/casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
@@ -23,4 +23,9 @@ public interface CasualConnection extends CasualServiceApi, CasualQueueApi, Casu
      */
     @Override
     void close();
+
+    DomainId getDomainID();
+
+    void addConnectionObserver(ConnectionObserver observer);
+
 }
--- a/casual/casual-api/src/main/java/se/laz/casual/jca/DomainId.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/jca/DomainId.java
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2022, The casual project. All rights reserved.
+ *
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+
 package se.laz.casual.jca;
 
 import java.io.Serializable;
--- a/casual/casual-api/src/main/java/se/laz/casual/jca/DomainId.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/jca/DomainId.java
@@ -1,0 +1,55 @@
+package se.laz.casual.jca;
+
+import java.io.Serializable;
+import java.util.Objects;
+import java.util.UUID;
+
+public class DomainId implements Serializable
+{
+    private static final long serialVersionUID = 1L;
+    private final UUID id;
+    private DomainId(UUID id)
+    {
+        this.id = id;
+    }
+
+    public static DomainId of(UUID id)
+    {
+        Objects.requireNonNull(id, "id can not be null");
+        return new DomainId(id);
+    }
+
+    public UUID getId()
+    {
+        return id;
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o)
+        {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass())
+        {
+            return false;
+        }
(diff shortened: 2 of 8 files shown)


aliothe:
@TobiasLeo @cklellie Any chance to get this reviewed before we're back from vacation?
At least then we "only" have the rewrite of casual caller to use this left to do to complete the issue.

cklellie:
I'll have a look now for you.

casual/casual-api/src/main/java/se/laz/casual/jca/DomainId.java
```diff
@@ -0,0 +1,61 @@
...
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+
+package se.laz.casual.jca;
+
+import java.io.Serializable;
+import java.util.Objects;
+import java.util.UUID;
+
+public class DomainId implements Serializable
```
cklellie, line 13: Do we really need a whole object to represent a just a UUID?

casual/casual-network/src/main/java/se/laz/casual/network/outbound/NettyNetworkConnection.java
```diff
@@ -72,10 +81,13 @@ public static NetworkConnection of(final NettyConnectionInformation ci, final Ne
         NettyNetworkConnection networkConnection = new NettyNetworkConnection(ci, correlator, ch, conversationMessageStorage, EventLoopFactory.getManagedExecutorService());
         LOG.finest(() -> networkConnection + " connected to: " + ci.getAddress());
         ch.closeFuture().addListener(f -> handleClose(networkConnection, networkListener));
-        networkConnection.throwIfProtocolVersionNotSupportedByEIS(ci.getProtocolVersion(), ci.getDomainId(), ci.getDomainName());
+        DomainId id = networkConnection.throwIfProtocolVersionNotSupportedByEIS(ci.getProtocolVersion(), ci.getDomainId(), ci.getDomainName());
```
cklellie, line 84: I am confused, `ci.getDomainId()`, is that not the same thing as we are being returned?

casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
```diff
@@ -23,4 +23,20 @@ public interface CasualConnection extends CasualServiceApi, CasualQueueApi, Casu
...
+    DomainId getDomainID();
+
+    /**
+     * If a connection is gone, for whatever reason - it is just not usable anymore.
+     *
+     * When this happens, {@link ConnectionObserver#connectionGone(DomainId)} is called with the same domain id
+     * that can be retrieved via {@link CasualConnection#getDomainID()}
+     * @param observer - the observer that will be notified
+     */
+    void addConnectionObserver(ConnectionObserver observer);
```
cklellie, line 40: Do we need to be able to remove observers too?

casual/casual-api/src/main/java/se/laz/casual/jca/DomainId.java
aliothe: Yes :)
The type is clearly what it is, a DomainId - so it does not matter what actual type is used internally, even if changed it still remains a DomainId.

casual/casual-network/src/main/java/se/laz/casual/network/outbound/NettyNetworkConnection.java
aliothe: The domain id is part of the handshake reply.

casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
aliothe: Not that I can see. A connection exists and can be used, by an observer, until notified that it can not be used anymore. That is the use case, it's gone thus both usage and observability.

casual/casual-network/src/main/java/se/laz/casual/network/outbound/NettyNetworkConnection.java
aliothe: One could argue that the method should be renamed to something like handshake perhaps?

cklellie: > The domain id is part of the handshake reply.

So what we can ask to connect to a domain id and another domain replies?

casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
cklellie: But who are the connection observers? Is this going to be casual caller?
Are the NettyNetworkConnection objects reused in the RA connection pool? If so, this could lead to the list of observers being populated with stale information as they are never cleared.
Symmetry gut feel, if you can request to observe something, you should also be able to request to no longer observe it too.

casual/casual-network/src/main/java/se/laz/casual/network/outbound/NettyNetworkConnection.java
aliothe: Each real network connetion only connects to one domain, during the handshake we get the domain id back in the reply.
For a pool of managed connections going to the same host:port it may be that the domain id for each managed connection is different. It can also be the same domain id - we do not know which.

casual/casual-api/src/main/java/se/laz/casual/jca/CasualConnection.java
aliothe: Connection observer will be anyone who wants to make use of a connection and who wants to keep caching information re the connection. This can only be done per domain id, thus why it has to be exposed. 
Casual caller will want want to use this to do its caching.

casual/casual-network/src/main/java/se/laz/casual/network/outbound/NettyNetworkConnection.java
aliothe: The problem we have today in such an application, casual caller for instance, is that it is not granular enough since it is caching based on connection factory jdni-name, that is - it expects all connections in a pool to be connected to the same domain - that is not necessarily true.

cklellie (approved):

PR #33, opened 2022-09-09
Feature/pooling backed by randomness

aliothe:
Since each network connection in the same pool can be connected to different casual insances, we have to take that into account.

Thus the caching is now based on the domain id for each connection that we get from the handshake with casual.

I am sorry that the PR is a bit too big but I had to rewrite casual-caller.

Re using JEE in JCA, that should in theory work fine but I could not get it to work on wildfly - thus singleton instead of ApplicationScoped and Injections.
Let's fix that issue on it's own in another feature.


aliothe:
Found an issue that need to be fixed and verified so closing this PR for now

PR #34, opened 2022-09-15
Feature/pooling backed by randomness

aliothe:
When connecting via a loadbalancer, each physical connection in a pool may be connected to different casual domains.
Thus they can each have their own set of queues and services and we need to take that into account.
We make it so that for each call we can provide the domain id so that the appserver can match on that and return a handle for which the physical connection has that exact domain id.

We record each domain id per host:port ( address) combination for all new ManagedConnections that the appserver creates.
This data can then be queried by casual-caller. 
Casual-caller does this by polling, by default each 10s, and then updating the state of the world.

The PR is a bit bigger than what we like.
The changes in the JCA part is not much with the exception of us now creating the physical connections right away instead of being lazy.
Casual-caller on the other hand is pretty much a rewrite - unfortunately.


casual/casual-jca/src/main/java/se/laz/casual/jca/CasualManagedConnectionFactory.java
```diff
@@ -27,19 +34,26 @@
...
 @SuppressWarnings("squid:S1948")
-public class CasualManagedConnectionFactory implements ManagedConnectionFactory, ResourceAdapterAssociation
+public class CasualManagedConnectionFactory implements ManagedConnectionFactory, ResourceAdapterAssociation, ValidatingManagedConnectionFactory
 {
-
-   private static final long serialVersionUID = 1L;
-   private static Logger log = Logger.getLogger(CasualManagedConnectionFactory.class.getName());
-   private ResourceAdapter ra;
-   private PrintWriter logwriter;
+    private static final long serialVersionUID = 1L;
```
TobiasLeo: Looks like incorrect indentation level here and in several other places in this file

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TpCallerImpl.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+        return makeServiceCall(cacheEntries, serviceName, callFunction);
+    }
+
+    private <R> R makeServiceCall(List<CacheEntry> matchingEntries, String serviceName, Function<CasualConnection, R> function)
+    {
+        List<CacheEntry> entries = matchingEntries.stream().collect(Collectors.toList());
+        Exception thrownException = null;
+        Collections.shuffle(entries);
```
TobiasLeo, line 95: Don't we lose the order by hops that was established earlier in the code? Unless I'm missing something this should only shuffle parts of the list with the same hops at once, not all of the list in one go

aliothe: The list only contains entries with the lowest hops.
We then shuffle to randomize where we issue the call.
https://github.com/casualcore/casual-java/blob/feature/pooling-backed-by-randomness/casual/casual-caller/src/main/java/se/laz/casual/connection/caller/CacheImpl.java#L116
Where the list comes from, only contains entries with the lowest hops and it is a new list.
There are unit tests that covers this.

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualManagedConnectionFactory.java
aliothe: Fixed

TobiasLeo (approved):
Looks good. Adding the concept of different resources behind a single IP does add complexity, but that's the world we need to handle

PR #36, opened 2022-10-03
Feature/network pooling

aliothe:
We provide the ability to configure the number of physical network connections per pool.
This is done by setting 2 additional attributes on the pool - NetworkPoolName and NetworkPoolSize.
This in addition to the old pool attributes, HostName and PortNumber.

If they are not configured then we still go with the 1-1 Managed Connection, physical network connection ratio.


casual/casual-jca/src/main/java/se/laz/casual/jca/CasualManagedConnection.java
```diff
@@ -84,20 +85,17 @@ public NetworkConnection getNetworkConnection()
...
-                                                                          .withDomainId(domain.getId())
-                                                                          .withDomainName(domain.getName())
-                                                                          .build();
-                networkConnection = NettyNetworkConnection.of(ci, this);
-                log.finest(()->"created new nw connection " + this);
+                if(null != mcf.getNetworkPoolName() && null == mcf.getNetworkPoolSize())
+                {
+                    log.warning(() -> "networkPoolName set to: " + mcf.getNetworkPoolName() + " but missing networkPoolSize!");
+                }
+                networkConnection = null != mcf.getNetworkPoolSize() && null != mcf.getNetworkPoolName() ? NetworkPoolHandler.getInstance().getOrCreate( mcf.getNetworkPoolName(), mcf.getAddress(), mcf.getCasualProtocolVersion(), this, mcf.getNetworkPoolSize()) : createOneToOneManagedConnection();
```
cklellie: this kind of tenary is not easy to read as it's too long, refactor with a private method returning optional<>, then use orElse to wrap the createOneToManagedConnection()  perhaps?

cklellie: refactor to make more readable.

TobiasLeo: This line almost reaches the 300 character mark. Maybe reformat for readability. For a ternary operator with an assignment I would think something like the following would help for line length and clarity
```
var = <boolean expression>
        ? <value if true>
        : <value else>
```

TobiasLeo (approved):
One the whole I think the pooling implementation looks good

outbound.md
```diff
@@ -48,3 +48,15 @@ Math.max(1, SystemPropertyUtil.getInt( "io.netty.eventLoopThreads", NettyRuntime
...
 If you are running on an older JVM in a containerized world, that might be suboptimal.
 
+## Pool configuration
+
+Without any specific configuration the mapping of Managed Connection entries, pool entries, and physical network connection is 1 to 1.
+You can specify the number of physical connections by specifying the network pool name and the number of connections you want.
+
+Example for wildfly:
+```
+$connectionDefinitionNode/config-properties=NetworkPoolName:add(value=pool-one)
```
cklellie: How about:
NetworkConnectionPoolName
NetworkConnectionPoolSize
Then it maps nicely with the code naming.

casual/casual-jca/src/main/java/se/laz/casual/jca/pool/NetworkConnectionPool.java
```diff
@@ -0,0 +1,193 @@
...
+                ", connections=" + connections +
+                ", poolName='" + poolName + '\'' +
+                ", poolSize=" + poolSize +
+                ", disconnected=" + disconnected +
+                '}';
+    }
+
+    private static ReferenceCountedNetworkConnection createNetworkConnection(Address address, ProtocolVersion protocolVersion, NetworkListener networkListener, ReferenceCountedNetworkCloseListener referenceCountedNetworkCloseListener, NetworkListener ownListener)
+    {
+        Domain domain = ConfigurationService.getInstance().getConfiguration().getDomain();
```
cklellie, line 128: code duplication with CasualManagedConnection#createOneToOneManagedConnection

casual/casual-jca/src/main/java/se/laz/casual/jca/jmx/CasualMBean.java
```diff
@@ -0,0 +1,9 @@
+package se.laz.casual.jca.jmx;
+
+import java.util.List;
+
+public interface CasualMBean
```
cklellie, line 5: So we just provide the pool names and the static count of what was specified for connections?
They should know that from how they deployed.
If we are going to start exposing JMX information we should determine some form of heirarchy etc. Otherwise it will not be very stable until we have added everything which is not nice for our end users.
Does this have to go with this PR?

casual/casual-jca/src/main/java/se/laz/casual/jca/pool/NetworkConnectionPool.java
```diff
@@ -0,0 +1,149 @@
...
+            throw new CasualResourceAdapterException("Address mismatch, have: " + this.address + " got: " + address + " for pool with name: " + poolName);
+        }
+        if(disconnected.get())
+        {
+            throw new CasualConnectionException("disconnected");
+        }
+        synchronized (getOrCreateLock)
+        {
+            // create up to pool size # of connections
+            // after that, randomly chose one - later on we can have some better heuristics for choosing which connection to return
```
cklellie: choose (present tense)

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/NetworkConnectionPoolTest.groovy
```diff
@@ -0,0 +1,78 @@
...
+{
+   def 'using the wrong address'()
+   {
+      given:
+      int poolSize = 5
+      def poolName = 'small-pool'
+      Address address = Address.of("nifty", 7771)
+      Address anotherAddress = Address.of('delta', 8787)
+      NetworkConnectionCreator connectionCreator = Mock(NetworkConnectionCreator){
+         1 * createNetworkConnection(*_) >> Mock(ReferenceCountedNetworkConnection)
```
cklellie: Use of wildcard matching on mock initialisation is not good practise. Should be more specific to also confirm that the calls are being called with the expected params.
Worst case here, the implementation could mess up returning a cache connection instance on first request, then on the wrong request it goes through, this test would not pick that up.
There are a number of other places you are doing the same with _ - again we should be checking the actual invokation parameters.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/NetworkConnectionPoolTest.groovy
```diff
@@ -0,0 +1,78 @@
...
+      connection != null
+      when: // using the wrong address - throws
+      pool.getOrCreateConnection(anotherAddress, protocolVersion, Mock(NetworkListener))
+      then:
+      thrown(CasualResourceAdapterException)
+   }
+
+   def 'size 1 always returns the same instance'()
+   {
+      int poolSize = 1
```
cklellie, line 43: given? setup?
Also what do we say about new lines before each block; when / then etc. We used to do this and it's much more readable in my opinion.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/ReferenceCountedNetworkConnectionTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+      def subsequentReferences = 4
+      def expectedNumberOfReferences = 5
+      def networkConnection = Mock(NettyNetworkConnection) {
+         1 * close()
+      }
+      def closeListener = Mock(ReferenceCountedNetworkCloseListener){
+         1 * closed(_)
+      }
+      // 1st reference count on creation
+      ReferenceCountedNetworkConnection instance = ReferenceCountedNetworkConnection.of(networkConnection, closeListener)
```
cklellie, line 46: This test sequential increments and decrements. Should we add a test using an Executor with multiple threads to ensure it's fine too?

casual/casual-jca/src/main/java/se/laz/casual/jca/jmx/CasualMBean.java
aliothe: It useful to expose runtime information, in this case regarding the new network pools.
I am using it myself while running stuff on wildfly.
Not sure what you mean by determining some form of hierarchy?

casual/casual-jca/src/main/java/se/laz/casual/jca/pool/NetworkConnectionPool.java
aliothe: Yeah we create actual network connections in both places.

outbound.md
aliothe: Fine with me, @TobiasLeo what do you say?

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/NetworkConnectionPoolTest.groovy
aliothe: yeah, that should be there for sure.
Didn't even know you could write spock tests without given or setup.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/ReferenceCountedNetworkConnectionTest.groovy
aliothe: Yeah, that won't hurt

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/NetworkConnectionPoolTest.groovy
aliothe: Please explain how it could mess up and returning a cached connection instance on first request due to this?
The connection creator is used inside the actual pool, in case a new ref counted connection should be created - that's what we're checking for, nothing else.

casual/casual-jca/src/main/java/se/laz/casual/jca/pool/NetworkConnectionPool.java
cklellie: yes - but it is the almost exact same snippet of code  - should we not have it in one place otherwise we might change something for creating with the network pool reference types and forget to change that in the non network pool reference later?

casual/casual-jca/src/main/java/se/laz/casual/jca/jmx/CasualMBean.java
cklellie: The data is just static data though - can't change after deployment scripts are run can it?

it's been a while since i played with JMX Beans, but i was thinking how it looks within the tree structure of jconsole - everything grouped in the tree appropriately based on function etc. So i thought do we want to take a beat and think what we want to have in general in the future and try and get these ones in the right place - or can we wait with these until later.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/NetworkConnectionPoolTest.groovy
cklellie: My point is that wildcards with mocks are not good for testing - as you aren't actually testing the dependency conditions effectively.
Here you don't validate that the connection request is recieving the correct address information.
The implementation could be changed to be something bananas - like the cache example i provided - which though insane, would still pass this test.
If however you validate the parameters of the dependent call - the test would fail when the connection is performed with the wrong address for example, causing a regression if someone breaks the implementation in the future.

aliothe: I agree that we can check the address and also the protocol version, no worries there.
The rest of the params would be troublesome, to say the least, to verify - those would also not add anything to what is being tested.

casual/casual-jca/src/main/java/se/laz/casual/jca/jmx/CasualMBean.java
aliothe: That data is very much alive when physical connections come and go.

aliothe: Re the grouping, that's namespace based.

aliothe: I can add a screen shot if you want.

casual/casual-jca/src/main/java/se/laz/casual/jca/pool/NetworkConnectionPool.java
aliothe: It's a usage of a builder in both cases, that's fine I think - unless you're thinking of something else?

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualManagedConnection.java
aliothe: Fixed

aliothe: Fixed

casual/casual-jca/src/main/java/se/laz/casual/jca/pool/NetworkConnectionPool.java
aliothe: Fixed

casual/casual-jca/src/test/groovy/se/laz/casual/jca/pool/NetworkConnectionPoolTest.groovy
aliothe: Fixed

outbound.md
TobiasLeo: Sounds good. It makes sense to use the term NetworkConnectionPool instead and it probably adds some clarity

PR #37, opened 2022-10-10
Feature/revert domain id

aliothe:
Revert domain id changes.
One pool is supposed to go towards one EIS only, if need be - by configuring only one physical connection.

casual-caller reverted to implementation from tag: 2.0.2 + using the current QueueInfo API.


TobiasLeo (approved):
Looks good to me. From what I can see all the old stuff is there and in order

PR #38, opened 2022-10-13
Feature/netty epoll

aliothe:
This enables us to choose if Netty should use NIO ( default) or epoll.

According to Netty's [documentation](https://netty.io/wiki/native-transports.html ) it may give us less garbage and better throughput so it makes sense to provide the option to use it.

To enable epoll, you either set CASUAL_OUTBOUND_USE_EPOLL to true or in the outbound section of your configuration file you add - ``` "useEpoll": true```.

Note, this does affect our packaging since now ```netty-transport-native-epoll``` will always be available, even when using NIO.
That in itself is not a problem.


TobiasLeo (approved):
Looks good

PR #39, opened 2022-10-19
Feature/jdk 8 11 17

aliothe:
Can use jdk 8, 11 or 17 to compile.
Still going with java 8 byte code generation.


TobiasLeo (approved):
Looks good. I tested it myself on jdk17 and build/test worked great. Nice to be able to build without problems on systems without old jdk versions

PR #42, opened 2022-10-24
Feature/reverse inbound

aliothe:
Reverse inbound, it is not a part of the JCA specification - however the opposite is available in casual, reverse outbound, so for symmetry reasons this feature is implemented.


TobiasLeo (approved):
Looks pretty solid. Very much non-jca, but exactly what we need and in line with casual

PR #45, opened 2022-11-03
Feature/casual caller handle tpenoent for cached entries

aliothe:
casual-caller

Scenario:
* There are cached service entries
* On tpcall for such an entry - TPENOENT is returned -> state of the world is not what we thought it was

Resolution:
* Remove the cached entry(ies) for that servicename
* Issue discovery again vs all valid pools
* Store in cache
* Get from cache as per usual
* Run tpcall again as per usual

This can happen if casual-caller is connected via a proxy and the casual instance(s) behind the proxy are restarted without the connection to casual-caller being cut.


casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
```diff
@@ -6,46 +6,86 @@
...
     {
-        Exception thrownException = null;
-
-        List<ConnectionFactoryEntry> prioritySortedFactories = lookup.get(serviceName);
+        List<ConnectionFactoryEntry> validEntries = getFoundAndValidEntries(lookup, serviceName);
 
-        // Service not found
-        if (prioritySortedFactories.isEmpty())
+        // No valid casual server found (revalidation is on a timer in ConnectionFactoryEntryValidationTimer)
+        if (validEntries.isEmpty())
```
cklellie, line 35: This code block is now duplicated in 3 places in this file.

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
```diff
@@ -6,46 +6,86 @@
...
 {
     private static final Logger LOG = Logger.getLogger(FailoverAlgorithm.class.getName());
 
-    public T callWithFailover(
+    public ServiceReturn<CasualBuffer> tpcallWithFailover(
             String serviceName,
             ConnectionFactoryLookup lookup,
-            FunctionThrowsResourceException<CasualConnection, T> doCall,
-            FunctionNoArg<T> doTpenoent)
+            Procedure serviceCacheRemover,
```
cklellie: Procedure is a very generic name. I am also not seeing the need to have everything as functions passed into methods. I know you miss CDI, but this is not the alternative imo. It only becomes slightly useful when you come to testing. If the Procedure was not a static function each time i could understand it. But in this case, it would be better to have it statically constructed in the FailOverAlorithm for example and just call it, you can always use constructor injection for testing if you wish to override and test the function is called, otherwise just assert on the side effects rather than mocking.

cklellie (approved):

PR #48, opened 2022-11-21
Feature/transaction pool sticky

TobiasLeo:


casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryLookup.java
```diff
@@ -36,8 +36,15 @@ public interface ConnectionFactoryLookup extends ServiceCacheMutator
...
     /**
      *
-     * @param serviceName - the name of he service
+     * @param serviceName - the name of the service
      * @return a List of 0-n CacheEntries
      */
     List<ConnectionFactoryEntry> get(String serviceName);
+
+    /**
+     * //TODO Documentation
```
aliothe: fix TODO

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
```diff
@@ -86,11 +87,27 @@ private List<ConnectionFactoryEntry> getFoundAndValidEntries(ConnectionFactoryLo
     private <T> T issueCall(String serviceName, List<ConnectionFactoryEntry> validEntries, FunctionThrowsResourceException<CasualConnection, T> doCall)
     {
         Exception thrownException = null;
+
+        // Sticky transaction handling
+        try
+        {
+            Optional<T> stickyMaybe = handleTransactionSticky(serviceName, validEntries, doCall);
+
+            if (stickyMaybe.isPresent()) return stickyMaybe.get();
```
aliothe: curlies, always - there are several conditional one liners where it is missing

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
```diff
@@ -114,6 +131,70 @@ private <T> T issueCall(String serviceName, List<ConnectionFactoryEntry> validEn
...
+            if (stickyFactoryMaybe.isPresent() && stickyFactoryMaybe.get().isValid())
+            {
+                // We have a specific stickied pool to use that looks usable, try to use it
+                ConnectionFactoryEntry connectionFactoryEntry = stickyFactoryMaybe.get();
+                factories.remove(connectionFactoryEntry); // If we later need to do failover stuff we don't want to retry with this one
+                LOG.finest(() -> "Attempting to use pool=" + connectionFactoryEntry.getJndiName() + " with sticky to current transaction.");
+
+                try (CasualConnection con = connectionFactoryEntry.getConnectionFactory().getConnection())
+                {
+                    if (con.serviceExists(serviceName))
```
aliothe: Does this mean that we always get extra lookups just because we're on the sticky path?
This even for currently known services?

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryEntryStore.java
```diff
@@ -6,6 +6,7 @@
 
 package se.laz.casual.connection.caller;
 
+import se.laz.casual.connection.caller.config.ConfigurationService;
```
aliothe, line 9: Sonar shows unused java.util.Optional import

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TransactionPoolMapper.java
```diff
@@ -0,0 +1,170 @@
...
+    {
+        this.transactionManager = transactionManager;
+    }
+
+    private TransactionPoolMapper()
+    {
+        // NOP
+    }
+
+    public static synchronized TransactionPoolMapper instance()
```
aliothe: We use getInstance in other places in the code base, would be good to keep with the same naming standard.

casual/casual-caller/src/test/groovy/se/laz/casual/connection/caller/TransactionPoolMapperTest.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+    def "Test with many transactions"()
+    {
+        setup:
+        int transactions = 5
+
+        for (int i = 0; i < transactions; i++)
+        {
+            transactionManager.setCurrentTransaction(new TransactionImpl(Status.STATUS_ACTIVE))
+            TransactionPoolMapper.instance().setPoolNameForCurrentTransaction("hello"+i)
```
aliothe: It should always be the same pool, pool name, for one transaction?

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
aliothe: lookup -> domain discovery towards casual

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TransactionPoolMapper.java
```diff
@@ -0,0 +1,176 @@
...
+        {
+            instance = new TransactionPoolMapper();
+        }
+
+        return instance;
+    }
+
+    public String getPoolNameForCurrentTransaction()
+    {
+        if (!stickyEnabled) return null;
```
aliothe: curlies

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TransactionPoolMapper.java
```diff
@@ -0,0 +1,176 @@
...
+
+        Optional<Transaction> transactionMaybe = getCurrentTransaction();
+        return transactionMaybe.map(transactionStickies::get).orElse(null);
+    }
+
+    public void setPoolNameForCurrentTransaction(String poolName)
+    {
+        if (!stickyEnabled) return;
```
aliothe: curlies

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
cklellie: Prefer to never use single line ifs, always enclose in `{` `}` to prevent mistakes later.

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
```diff
@@ -114,6 +131,64 @@ private <T> T issueCall(String serviceName, List<ConnectionFactoryEntry> validEn
...
+     * @return Optional ServiceReturn. An empty result could indicate that stickiness isn't enabled, sticky isn't set yet for the current transaction (which will be the case for the first call) or the stickied pool was unavailable so call to sticky was skipped. Empty should always lead to retry down the line if possible, otherwise a TPENOENT response.
+     * @throws ResourceException Some softer errors are reported as resource exceptions. If these are thrown later retries with other pools is possible.
+     */
+    private <T> Optional<T> handleTransactionSticky(
+            String serviceName,
+            List<ConnectionFactoryEntry> factories,
+            FunctionThrowsResourceException<CasualConnection, T> doCall) throws ResourceException
+    {
+        if (TransactionPoolMapper.getInstance().isPoolMappingActive())
+        {
```
cklellie, line 152: prefer shortcut returns on negation to reduce large if blocks.
e.g.  
```
if( ! TransactionPoolMapper.getInstance().isPoolMappingActive() )
{
    return Optional.empty();
} 
```

casual/casual-caller/src/test/groovy/se/laz/casual/connection/caller/TransactionPoolMapperTest.groovy
```diff
@@ -0,0 +1,70 @@
...
+
+        String actualPoolName = "hello, world!"
+        TransactionPoolMapper.getInstance().setPoolNameForCurrentTransaction(actualPoolName)
+
+        expect:
+        TransactionPoolMapper.getInstance().getPoolNameForCurrentTransaction() == actualPoolName
+        TransactionPoolMapper.getInstance().getNumberOfTrackedTransactions() == 1
+    }
+
+    def "Test with many transactions"()
```
cklellie: Am I correct is saying that, this all relies on the Transaction implementations in wildfly and weblogic implementing equals and hashcode controls which actually utilise the transaction id somehow?
If so can we add tests which use those actual objects instead - include them as deps to testCompile for example, then we can actually test with those objects.

aliothe (changes requested):

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
TobiasLeo: I'll change it. There are a few places where I've been a bit sloppy with this

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
TobiasLeo: You are correct, that makes for a nicer method. Deep nesting and very wide code blocks get nasty fast

casual/casual-caller/src/test/groovy/se/laz/casual/connection/caller/TransactionPoolMapperTest.groovy
TobiasLeo: Correct, the equality holds true for both and we need it to work like this. It should be possible to test with the actual Transaction-classes for WildFly at least where I should be able to pull in some Narayana dependency. Don't know about if I can get my hands on implementation classes  for WLS in some convenient manner, though. I'll have to look into it

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryLookup.java
TobiasLeo: I thought I deleted that, it was only used in an earlier iteration of the solution. I'll just delete it

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/ConnectionFactoryEntryStore.java
TobiasLeo: I'll remove it. Was left over from old config implementation

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TransactionPoolMapper.java
TobiasLeo: Fixed

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TransactionPoolMapper.java
TobiasLeo: Fixed

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/TransactionPoolMapper.java
TobiasLeo: Fixed

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
TobiasLeo: This was probably used at some point, but after some rewrites it was not needed. Has been removed

casual/casual-caller/src/main/java/se/laz/casual/connection/caller/FailoverAlgorithm.java
TobiasLeo: Fixed

casual/casual-caller/src/test/groovy/se/laz/casual/connection/caller/TransactionPoolMapperTest.groovy
TobiasLeo: Should always be the same for the same transaction, but in this case we are using different transactions to add different pools just to see that we get a correct number of tracked transactions in the end. For the purpose of the test using different pool names for each transaction has no real bearing though since the poolmapper does not care if 5 different transactions stick to the same or different pools

aliothe: Sticky with the same transaction but different pools would be an error though currently, no?

TobiasLeo: Normally we should never set a sticky for a transaction where it is already set. Technically you could do it and it would result in the stickied pool being changed to the last pool that was set for the given transaction. If the test in question used the same transaction five times it would be expected in the end that we would be tracking one stickied transaction instead of five with how it is implemented

aliothe: I'm reading this with the eyes of somebody who does not know how it is supposed to work.
That is, that we only ever sticky a transaction to one pool only - anything else is not ok.
This might confuse that?

casual/casual-caller/src/test/groovy/se/laz/casual/connection/caller/TransactionPoolMapperTest.groovy
TobiasLeo: After looking at this it may be too much of a mess to build any useful tests on the WildFly specific classes. It quickly starts to devolve into Narayana implementation detail, different wrapper classes and so on. I'm not really sure these tests would be well invested time after gaining some more knowledge

aliothe (commented):
Fix the fix

casual/casual-caller/src/test/groovy/se/laz/casual/connection/caller/TransactionPoolMapperTest.groovy
TobiasLeo: I made the test clearer and added a new test to verify a more strict behavior for when multiple attempts are made to set the sticky for the same transaction

PR #52, opened 2022-12-16
log start inbound exception as SEVERE

aliothe:
Exceptions when starting inbound should be logged as SEVERE not WARNING


TobiasLeo (approved):
Makes sense, if inbound throws on startup it probably needs attention

PR #53, opened 2023-01-10
Feature/optional inbound delay start of server

aliothe:
This is an optional setting to help with a problem that we've only seen on wls.
When starting the appserver, during the call to endPointActivation on the resource adapter, if there are incoming domain discoveries it may happen that we do not get a handle via the MessageEndpointFactory.
This then throws, which is according to the specification.
However, it never self heals - that is, the activation never completes.

With this optional setting a user can set an initial so that we delay the start of the inbound server.
The endpointActivation call from the appserver runs in another thread and should thus always be able to become activated and not trip over itself.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
@@ -13,13 +13,17 @@ public class Inbound
 {
     public static final String CASUAL_INBOUND_STARTUP_MODE = "CASUAL_INBOUND_STARTUP_MODE";
     public static final String CASUAL_INBOUND_USE_EPOLL = "CASUAL_INBOUND_USE_EPOLL";
+    public static final String CASUAL_INBOUND_INITIAL_DELAY_ENV_NAME = "CASUAL_INBOUND_INITIAL_DELAY";
+
     private final Startup startup;
     private final boolean useEpoll;
+    private final long initialDelay;
 
     public Inbound( Builder builder )
     {
         this.startup = builder.startup;
         this.useEpoll = builder.useEpoll;
+        this.initialDelay = builder.initialDelay;
     }
 
     public Startup getStartup()
@@ -30,6 +34,11 @@ public class Inbound
     public boolean isUseEpoll()
     {
         return useEpoll;
+    }
+
+    public long getInitialDelay()
+    {
+        return initialDelay;
     }
 
     @Override
@@ -71,6 +80,7 @@ public class Inbound
     {
         private Startup startup;
         private Boolean useEpoll;
+        private Long initialDelay;
 
         private Builder()
         {
@@ -85,6 +95,12 @@ public class Inbound
         public Builder withStartup( Startup startup )
         {
             this.startup = startup;
+            return this;
+        }
+
+        public Builder withInitialDelay(long initialDelay)
+        {
+            this.initialDelay = initialDelay;
             return this;
         }
 
@@ -103,6 +119,12 @@ public class Inbound
             if( useEpoll == null )
             {
                 useEpoll = Boolean.parseBoolean( Optional.ofNullable( System.getenv( CASUAL_INBOUND_USE_EPOLL ) ).orElse( "false" ) );
+            }
+            if( initialDelay == null )
+            {
+                initialDelay = Optional.ofNullable( System.getenv( CASUAL_INBOUND_INITIAL_DELAY_ENV_NAME ) )
+                                       .map( delay -> delay.isEmpty() ? 0L : Long.parseLong( delay ) )
+                                       .orElse(0L );
             }
             return new Inbound( this );
         }
--- a/casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
@@ -13,7 +13,7 @@ public class Inbound
 {
     public static final String CASUAL_INBOUND_STARTUP_MODE = "CASUAL_INBOUND_STARTUP_MODE";
     public static final String CASUAL_INBOUND_USE_EPOLL = "CASUAL_INBOUND_USE_EPOLL";
-    public static final String CASUAL_INBOUND_INITIAL_DELAY_ENV_NAME = "CASUAL_INBOUND_INITIAL_DELAY";
+    public static final String CASUAL_INBOUND_STARTUP_INITIAL_DELAY_ENV_NAME = "CASUAL_INBOUND_STARTUP_INITIAL_DELAY_IN_SECONDS";
 
     private final Startup startup;
     private final boolean useEpoll;
@@ -53,13 +53,14 @@ public class Inbound
             return false;
         }
         Inbound inbound = (Inbound) o;
-        return isUseEpoll() == inbound.isUseEpoll() && Objects.equals( getStartup(), inbound.getStartup() );
+        return isUseEpoll() == inbound.isUseEpoll() && Objects.equals( getStartup(), inbound.getStartup() ) &&
+                inbound.getInitialDelay() == getInitialDelay();
     }
 
     @Override
     public int hashCode()
     {
-        return Objects.hash( getStartup(), isUseEpoll() );
+        return Objects.hash( getStartup(), isUseEpoll(), getInitialDelay() );
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
```diff
@@ -13,13 +13,17 @@
 {
     public static final String CASUAL_INBOUND_STARTUP_MODE = "CASUAL_INBOUND_STARTUP_MODE";
     public static final String CASUAL_INBOUND_USE_EPOLL = "CASUAL_INBOUND_USE_EPOLL";
+    public static final String CASUAL_INBOUND_INITIAL_DELAY_ENV_NAME = "CASUAL_INBOUND_INITIAL_DELAY";
```
cklellie: `CASUAL_INBOUND_STARTUP_DELAY` or `CASUAL_INBOUND_STARTUP_INITIAL_DELAY` - to be consistent with existing `CASUAL_INBOUND_STARTUP_MODE`

casual/casual-jca/src/main/java/se/laz/casual/jca/work/Delayer.java
```diff
@@ -0,0 +1,36 @@
...
+        try
+        {
+            log.finest(() -> "delaying current thread by " + seconds + " seconds");
+            Duration delay = Duration.ofSeconds(seconds);
+            TimeUnit.SECONDS.sleep(delay.get(ChronoUnit.SECONDS));
+        }
+        catch (InterruptedException e)
+        {
+            Thread.currentThread().interrupt();
+            throw new InboundStartupException("Interrupted while delaying inbound startup", e);
```
cklellie: Delayer is generic - but exception and message are not?

configuration.md
```diff
@@ -13,6 +13,7 @@ Casual supports the following configuration options which can be set using envir
     See [Inbound Startup Configuration](inbound.md#startup-configuration) for more details.
 * `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
 * `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
+* `CASUAL_INBOUND_INITIAL_DELAY` - Delay the startup of the inbound server. Default is no delay.
```
TobiasLeo: In what unit of measure? Seconds, milliseconds?

casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
```diff
@@ -96,6 +104,23 @@ private Set<String> checkRemainingServices( Set<String> remaining )
...
 
+    private void maybeDelay()
+    {
+        // Note:
+        // This, optional, delay of inbound is due to an issue on wls where once in a blue moon - the MessageEndpointFactory is in fact not ready
+        // and there are incoming domain discoveries ( during restart)
+        // That will then fail and continue to fail forever, even after endpointActivation for the resource adapter has completed - it should heal
+        // This is a horrible "work around" for that specific problem - the user decides how long inbound startup has to be delayed
+        // Never seen on wildfly
+        if(delay <= 0)
```
cklellie: `0L`

configuration.md
```diff
@@ -13,6 +13,7 @@ Casual supports the following configuration options which can be set using envir
...
 
 NB - if a `CASUAL_CONFIG_FILE` is provided, it take precedence over the `CASUAL_INBOUND_STARTUP_MODE` setting.
```
cklellie, line 18: Add note in the documentation about "when" the delay is implemented in the case of startup mode triggered. I.e. we wait for triggered application to complete and then delay. Otherway around could be confusing as you might think the delay and the wait run concurrently.

casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
```diff
@@ -32,6 +36,11 @@ public boolean isUseEpoll()
...
     }
 
+    public long getInitialDelay()
+    {
+        return initialDelay;
+    }
+
     @Override
     public boolean equals( Object o )
     {
```
TobiasLeo, line 46: Can't add a comment down at toString, but it should include the new `initialDelay`

configuration.md
aliothe: Bananas?
I'll add seconds :)

casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
```diff
@@ -13,7 +13,7 @@
 {
     public static final String CASUAL_INBOUND_STARTUP_MODE = "CASUAL_INBOUND_STARTUP_MODE";
     public static final String CASUAL_INBOUND_USE_EPOLL = "CASUAL_INBOUND_USE_EPOLL";
-    public static final String CASUAL_INBOUND_INITIAL_DELAY_ENV_NAME = "CASUAL_INBOUND_INITIAL_DELAY";
+    public static final String CASUAL_INBOUND_STARTUP_INITIAL_DELAY_ENV_NAME = "CASUAL_INBOUND_STARTUP_INITIAL_DELAY_IN_SECONDS";
```
cklellie: `_IN_SECONDS` feels OTT. What do you think @TobiasLeo 
Eitherway consistency in the const naming would be nice.

configuration.md
```diff
@@ -13,10 +13,14 @@ Casual supports the following configuration options which can be set using envir
     See [Inbound Startup Configuration](inbound.md#startup-configuration) for more details.
 * `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
 * `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
+* `CASUAL_INBOUND_STARTUP_INITIAL_DELAY_IN_SECONDS` - Delay the startup of the inbound server. Default is no delay.
 
 NB - if a `CASUAL_CONFIG_FILE` is provided, it take precedence over the `CASUAL_INBOUND_STARTUP_MODE` setting.
 However, if some configuration is missing from the configuration file but has a setting via an environment variable then this will be used before any hardcoded default setting.
 
+If `CASUAL_INBOUND_STARTUP_INITIAL_DELAY_IN_SECONDS` is configured then it will always delay the startup by that amount of seconds regardless of startup mode.
```
cklellie: Need to update the Inbound.md too, so as to make this clear in the different modes what happens.
i.e. IMMEDIATE actually with this setting will just immediately start waiting for the delay to complete.
TRIGGER will wait until the trigger app is registered and then delay.
DISCOVERY will wait until all the registered inbound services are found and then delay.
It's the first one which is the bad apple as it is contradictory - IMMEDIATE - but not quite as you want to delay! :|

casual/casual-api/src/main/java/se/laz/casual/config/Inbound.java
aliothe: @TobiasLeo Think of a person, not the one who configured it, who reads the env var and the value.
With IN_SECONDS, they know what it is - without it, they'd have to guess or go read some other documentation.
It is a bit long but I think the information conveyed makes it well worth it, less error prone all around.

TobiasLeo: I think it is useful to have  the unit stated in the env if it is fixed. Any user stumbling upon it will understand the value more easily without having to look at the documentation or source code. Personally I would drop the IN though as I feel it is kind of redundant. I often feel like having <some time related task>_<unit> to be sufficient to tell the user what is configured and what the number will be interpreted as

I agree with @cklellie about naming consistency, either they should all be suffixed with _ENV_NAME or none of them should have the suffix

aliothe: I'll drop the IN part.
Re ENV_NAME for constants that are names for env vars, if you check the rest of the code base it is with ENV_NAME since that conveys the meaning of that the constant is.
Thus. the other ones, if they are indeed env vars, should be changed - not in this PR though.
I'll create an issue to go over ALL constants in the code and make sure that env vars do have ENV_NAME in them ( I think 99% do already have that).

aliothe: https://github.com/casualcore/casual-java/issues/54

TobiasLeo (approved):

PR #61, opened 2023-02-06
Feature/stop jndi search timer if trigger mode and inbound started

aliothe:
If inbound mode is trigger and inbound has started, there is no need for the JNDI search timer to continue running.
This since when running in trigger mode, no dynamic deployments are expected.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/config/Startup.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/config/Startup.java
@@ -16,9 +16,9 @@ import java.util.logging.Logger;
 public class Startup
 {
     private static final Logger LOG = Logger.getLogger(Startup.class.getName());
-    private static final String CASUAL_INBOUND_STARTUP_MODE_ENV_NAME = "CASUAL_INBOUND_STARTUP_MODE";
     private final Mode mode;
     private final List<String> services;
+    public static final String CASUAL_INBOUND_STARTUP_MODE_ENV_NAME = "CASUAL_INBOUND_STARTUP_MODE";
 
     public Startup( Mode mode, List<String> services )
     {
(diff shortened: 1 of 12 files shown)


casual/casual-api/src/main/java/se/laz/casual/config/Startup.java
```diff
@@ -101,6 +103,7 @@ public Builder withServices( List<String> services )
 
         public Startup build()
         {
+            mode = (null == mode) ? Mode.fromName(Optional.ofNullable(System.getenv(CASUAL_INBOUND_STARTUP_MODE_ENV_NAME)).orElseGet(Mode.IMMEDIATE::getName)) : mode;
```
TobiasLeo: We already have ConfigurationService that can and should serve this config. I can see though that the code calling this from Inbound is some "older" code not touched by this PR. Also, wouldn't this fail to read the inbound startup mode if it is configured by file?

casual/casual-inbound-handler-casual-service/build.gradle
```diff
@@ -19,9 +19,11 @@ dependencies {
   implementation project(":casual:casual-inbound-handler-api")
   implementation project(":casual:casual-service-discovery-extension")
   compileOnly libs.javaee_api
-
+  compileOnly libs.gson
+    
   testImplementation libs.javaee_api
   testImplementation libs.gson
+  testImplementation libs.system_lambda
```
TobiasLeo, line 26: Is system lambda actually ever used? I can't see it in any tests in this version, and seems to work fine locally even if I remove it

casual/casual-api/src/main/java/se/laz/casual/config/Startup.java
aliothe: Fixed, going with config files and also testing using config files and overriding the inbound startup mode via env var.

casual/casual-inbound-handler-casual-service/build.gradle
aliothe: It was earlier and now it is again :)

TobiasLeo (approved):

PR #62, opened 2023-03-03
Feature/maven central publishing#17

cklellie:
I have left a few things commented out so you can play with it yourself first.

The majority of the changes are to fix javadocs that break the build with werror enabled not that we generate javadocs. Though i disabled the java 17 checks as these were very strict!

Then it is new gradle version, not the latest as this breaks the tests using system lambda.

Then a whole lot of changes to the dependency setup to ensure things are correct in the published pom to maven central.


aliothe (approved):
Works fine, tested by publishing a SNAPSHOT version

TobiasLeo:
Seems to work fine once configured as described. Two questions:
Why is casual-caller publishing disabled?
With the changes that introduce signing I was unable to even build without first adding a dummy config with all the keys for the signing plugin. That feels sort of sub-optimal. Is there any way to circumvent this? Preferably I should only need to add those gradle properties in places where I actually have my signing keys and have intent to sign

cklellie:
> 

Casual-caller is going to be moved into a seperate repository where it can live with it's own version tag etc. It was otherwise confusing trying to publish things with different coordinates.

It is intended that it should still be able to build from a clean clone, i had tested multiple times along the way, but perhaps some of the last bits of config have broken that. I'll have a check a bit later and see what is wrong.
Which command do you run, just `./gradlew clean build` ?
The only command that should fail without those configurations should be `./gradlew publish`

TobiasLeo:
Ok, moving casual caller to be able to better differentiate and tag things sounds good

I'm still getting issues with a clean, unconfigured repository still, though. If I just do a fresh clone and then try to build from this branch I get the same issue on two different computers I've tested: one linux host (openjdk17) and one Mac (openjdk8). Seems any gradle command yields the same issue

```
$ ./gradlew clean build

FAILURE: Build failed with an exception.

* Where:
Script '/home/tobias/git/casual-java/publish.gradle' line: 70

* What went wrong:
A problem occurred evaluating script.
> Could not get unknown property 'ossrhUsername' for Credentials [username: null] of type org.gradle.internal.credentials.DefaultPasswordCredentials_Decorated.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 890ms
```

Don't know if I'm doing anything weird, or if you happen to have some global ~/.gradle/gradle.properties or something that has the config? Looks like we're trying to do a raw property read on publish.gradle:70, it may need to be wrapped with a check if the property exists (along with the following line)

I don't know what might be the cleanest solution, but something like the following for lines 70-71 seems to fix the issue for me
```groovy
                username = project.hasProperty('ossrhUsername') ? ossrhUsername : null
                password = project.hasProperty('ossrhPassword') ? ossrhPassword : null
```

cklellie:
> Ok, moving casual caller to be able to better differentiate and tag things sounds good
> 
> I'm still getting issues with a clean, unconfigured repository still, though. If I just do a fresh clone and then try to build from this branch I get the same issue on two different computers I've tested: one linux host (openjdk17) and one Mac (openjdk8). Seems any gradle command yields the same issue
> 
> ```
> $ ./gradlew clean build
> 
> FAILURE: Build failed with an exception.
> 
> * Where:
> Script '/home/tobias/git/casual-java/publish.gradle' line: 70
> 
> * What went wrong:
> A problem occurred evaluating script.
> > Could not get unknown property 'ossrhUsername' for Credentials [username: null] of type org.gradle.internal.credentials.DefaultPasswordCredentials_Decorated.
> 
> * Try:
> > Run with --stacktrace option to get the stack trace.
> > Run with --info or --debug option to get more log output.
> > Run with --scan to get full insights.
> 
> * Get more help at https://help.gradle.org
> 
> BUILD FAILED in 890ms
> ```
> 
> Don't know if I'm doing anything weird, or if you happen to have some global ~/.gradle/gradle.properties or something that has the config? Looks like we're trying to do a raw property read on publish.gradle:70, it may need to be wrapped with a check if the property exists (along with the following line)
> 
> I don't know what might be the cleanest solution, but something like the following for lines 70-71 seems to fix the issue for me
> 
> ```groovy
>                 username = project.hasProperty('ossrhUsername') ? ossrhUsername : null
>                 password = project.hasProperty('ossrhPassword') ? ossrhPassword : null
> ```

I can reproduce too - looks like i had a gradle.properties file still there, git reset didn't remove it as it was in the .gitignore. Your suggesting seems good to me. I'll take it for a test drive.

TobiasLeo (approved):

PR #64, opened 2023-03-22
Feature/extension refactoring

cklellie:
Updated extensions to be more generic such that they can work with any type of ServiceHandler, as long as the handler integrates them into the invokeService method.
Therefore dropped the `Casual` from the naming to reflect this.
I still feel that the naming could more accurately described as `ServiceInvocationExtension` or `ServiceHandlerInvocationExtension` as that is more what it is, you can't extend anything but the `invokeService` call. But i have not changed this.
I have however changed the name of `convert` to `convertRequestParams` as it was ambigious as to it's intention.
With `handleError` this has changed such that the `ServiceHandler` can build an initial response and the extension can then modify this prior to returning, which feels simpler.
Because of that change, there is also now no need for a default implementation specifically to casual, or javaee service handler, as the default interface implementation take care of the rest. The only non defaulted method is `canHandle` - which is to ensure those creating extensions remember to set it appropriately and don't accidentally take over all service calls.


casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/service/extension/ServiceHandlerExtensionFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Retrieve the most appropriate {@link ServiceHandlerExtension} base on it's {@link Priority}.
+     *
+     * Default extension {@link DefaultServiceHandlerExtension} is always registered with the lowest priority.
+     *
+     * @param name provided by the service handler to filter extensions based upon.
+     * @return the service handler extension.
+     */
+    public static ServiceHandlerExtension getExtension( String name )
```
aliothe, line 47: We are now potentially returning null here.
If the return value is Optional<ServiceHandlerExtension> then that should be shown in the api so that a service handler may handle it in it's own code and we'll not suffer a needless null pointer exception.

aliothe: Ofc, that should never happen due to the default handler being loaded via SPI.
Still, the code reads as if it might happen.
Add a post condition that the found being returned is not null - for readability?

aliothe (approved):

PR #65, opened 2023-03-28
wls needs Stateless annotation

aliothe:
It works fine on wildfly without the Stateless annotations but for wls that is not so.


cklellie (commented):
You have added back the @Stateless but not the @Local annotations that were also removed. Have you tested to confirm that this works now with just the @Stateless?

aliothe:
Yes, I have tested it on wls as well now - making inbound calls.

TobiasLeo (approved):

PR #66, opened 2023-04-04
casual caller external

aliothe:
casual caller is now external and lives here:
https://github.com/casualcore/casual-caller


cklellie (commented):
You have missed removal of the casual caller exclude for publishing in `/build.gradle` line 49.

cklellie (approved):

PR #67, opened 2023-04-05
name change of SPI Extension file

aliothe:


aliothe:
Verified using quay.io/wildfly/wildfly:27.0.0.Final-jdk17

cklellie (approved):

PR #72, opened 2023-05-16
Feature/2.2/inbound log waiting for service registration

aliothe:
If the inbound startup mode is discovery, log the initial set of services and keep logging which service registrations that inbound is still waiting for.
This is due to a demand from a customer and it is not an unreasonable request.

### DIFF
--- a/casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
+++ b/casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
@@ -80,9 +80,9 @@ public final class StartInboundServerWork<T> implements Work
         Set<String> remaining = checkRemainingServices( new HashSet<>( this.startupServices ) );
         while(!remaining.isEmpty())
         {
-            int previouslyNumberOfServicesRemaining = remaining.size();
+            int previousNumberOfServicesRemaining = remaining.size();
             remaining = checkRemainingServices( remaining );
-            if(!remaining.isEmpty() && remaining.size() < previouslyNumberOfServicesRemaining)
+            if(!remaining.isEmpty() && remaining.size() < previousNumberOfServicesRemaining)
             {
                 logWaitingForStartupServices(remaining);
             }
--- a/casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
+++ b/casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
@@ -79,11 +79,14 @@ public final class StartInboundServerWork<T> implements Work
     {
         logInitialStartupServices(startupServices);
         Set<String> remaining = checkRemainingServices( new HashSet<>( this.startupServices ) );
-        logWaitingForStartupServices(remaining);
         while(!remaining.isEmpty())
         {
+            int remainingServices = remaining.size();
             remaining = checkRemainingServices( remaining );
-            logWaitingForStartupServices(remaining);
+            if(!remaining.isEmpty() && remaining.size() < remainingServices)
+            {
+                logWaitingForStartupServices(remaining);
+            }
             try
             {
                 Thread.sleep( 1000 );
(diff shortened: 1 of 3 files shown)


casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
```diff
@@ -74,11 +77,13 @@ public void run()
...
     private void waitForInboundStartupServices()
     {
-        log.info(() -> "Waiting for " + startupServices.size() + " startup services to be registered before inbound starts.");
+        logInitialStartupServices(startupServices);
         Set<String> remaining = checkRemainingServices( new HashSet<>( this.startupServices ) );
+        logWaitingForStartupServices(remaining);
         while(!remaining.isEmpty())
         {
             remaining = checkRemainingServices( remaining );
+            logWaitingForStartupServices(remaining);
```
cklellie: only log if there is a change in remaining to prevent spamming the log.

casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
```diff
@@ -79,11 +79,14 @@ private void waitForInboundStartupServices()
     {
         logInitialStartupServices(startupServices);
         Set<String> remaining = checkRemainingServices( new HashSet<>( this.startupServices ) );
-        logWaitingForStartupServices(remaining);
         while(!remaining.isEmpty())
         {
+            int remainingServices = remaining.size();
```
cklellie: rename `previousCount`?

casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
```diff
@@ -92,6 +100,25 @@ private void waitForInboundStartupServices()
         log.info(() -> "All startup services registered.");
     }
 
+    private void logInitialStartupServices(List<String> startupServices)
+    {
+        if(ConfigurationService.getInstance().getConfiguration().getInbound().getStartup().getMode() == Mode.DISCOVER)
+        {
+            log.info(() -> "Waiting for " + startupServices.size() + " services to be registered before inbound starts.");
```
cklellie: This is a change from the current logging behaviour, previously we always logged this line.

aliothe: Yes it does.
However, does that logging make any sense if startup mode is not discover?
That is, is it a valid case to have a list of startup services with any other mode than discover?
If so, what are they?

cklellie: Trigger mode - in that case there is 1 service it is waiting for, this would remove the logging showing it is waiting for that.
 
 https://github.com/casualcore/casual-java/blob/dev/casual/casual-api/src/main/java/se/laz/casual/config/Startup.java#L33-L44
 
Perhaps if we add a line saying "Inbound startup mode X".
Then remove the checks for which mode for the logging as they make sense.

```
Inbound startup mode IMMEDIATE
Waiting for 0 services...
All startup services registered.
```
```
Inbound startup mode TRIGGER
Waiting for 1 services...
Name...
All startup services registered.
```
```
Inbound startup mode DISCOVER
Waiting for x services...
Name...
Waiting for x-2 services...
Name...
All startup services registered.
```

casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
```diff
@@ -74,11 +76,16 @@ public void run()
 
     private void waitForInboundStartupServices()
     {
-        log.info(() -> "Waiting for " + startupServices.size() + " startup services to be registered before inbound starts.");
+        logInitialStartupServices(startupServices);
         Set<String> remaining = checkRemainingServices( new HashSet<>( this.startupServices ) );
         while(!remaining.isEmpty())
         {
+            int previouslyNumberOfServicesRemaining = remaining.size();
```
cklellie: grammar - `previous` not `previously`

cklellie (approved):

PR #76, opened 2023-06-08
Feature/2.2/casual protocol 1.1 outbound

aliothe:
Handle domain disconnect request by casual.
Note, this is for outbound connections only.


casual/casual-api/src/main/java/se/laz/casual/network/connection/DomainDisconnectedException.java
TobiasLeo, line 1: Missing license header

TobiasLeo (approved):

PR #78, opened 2023-08-30
Feature/2.2/topology update

aliothe:
Bump and support protocol version 1.2
It adds a topology updated event that we expose via ```CasualConnection`.

It means that the services known for that connection has changed in some way.
This is useful information for any application that caches such data.


casual/casual-network/src/main/java/se/laz/casual/network/outbound/NettyNetworkConnection.java
```diff
@@ -92,11 +96,16 @@ public static NetworkConnection of(final NettyConnectionInformation ci, final Ne
         ch.closeFuture().addListener(f -> handleClose(networkConnection, errorInformer));
         DomainId id = networkConnection.throwIfProtocolVersionNotSupportedByEIS(ci.getDomainId(), ci.getDomainName());
         networkConnection.setDomainId(id);
-        if(networkConnection.getProtocolVersion() == ProtocolVersion.VERSION_1_1 || networkConnection.getProtocolVersion() == ProtocolVersion.VERSION_1_2)
+        if(networkConnection.isProtocolVersionOneOneOrOneTwo())
```
TobiasLeo: Could `networkConnection.isProtocolVersionOneOneOrOneTwo()` and `networkConnection.isProtocolVersionOneTwo()` be broken up into specific feature checks, like `networkConnection.protocolSupportsDomainDisconnect()` and `networkConnection.protocolSupportsDomainTopologyChange()`. If casual adds more protocol levels methods don't need renaming and we won't get methods like isProtocolVersionOneOneOrOneTwoOrOneThree and so on

aliothe: Done

TobiasLeo (approved):
Looks good

PR #80, opened 2023-09-13
Workaround/2.2.22/wls prepare xa terminator deadlock

aliothe:
Workaround for deadlock in XATerminator::prepare on weblogic 12c.
We are not sure that there are any side effects of this fix since we conditionally skip using XATerminator::prepare and are thus not adhering to the JCA specification anymore.
Beware!

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/jca/GlobalTransactionIdHandler.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/jca/GlobalTransactionIdHandler.java
@@ -29,6 +29,7 @@ public final class GlobalTransactionIdHandler
     public void addGtrid(GlobalTransactionId globalTransactionId, DomainId belongsTo)
     {
         Objects.requireNonNull(globalTransactionId, GLOBAL_TRANSACTION_ID_CAN_NOT_BE_NULL);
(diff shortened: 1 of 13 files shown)


TobiasLeo (approved):
Looks good to me

casual/casual-api/src/main/java/se/laz/casual/jca/GlobalTransactionIdHandler.java
```diff
@@ -0,0 +1,71 @@
...
+    {}
+
+    public static GlobalTransactionIdHandler of()
+    {
+        return new GlobalTransactionIdHandler();
+    }
+
+    public void addGtrid(GlobalTransactionId globalTransactionId, DomainId belongsTo)
+    {
+        Objects.requireNonNull(globalTransactionId, GLOBAL_TRANSACTION_ID_CAN_NOT_BE_NULL);
```
TobiasLeo, line 31: Shouldn't `DomainId belongsTo` also be requiredNonNull here?

aliothe: Indeed

TobiasLeo (approved):

PR #81, opened 2023-10-12
buggfix/3.2/pass real method to buffer handlers

aliothe:
The real method needs to be available for buffer handlers since annotations may only exist there and not on the proxy method.
We also future proof the API by using a wrapper class for the inbound request info that can be further expanded upon when need be.

Other changes:
* for spock using Java 17 
* using bytebuddy instead of cglib since it gives better error messages when things go wrong
* up version of netty to 4.1.100


casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/buffer/PassThroughBufferHandler.java
```diff
@@ -30,23 +29,23 @@ public boolean canHandleBuffer(String bufferType)
     }
 
     @Override
-    public ServiceCallInfo fromRequest(Proxy p, Method method, InboundRequest request)
+    public ServiceCallInfo fromRequest(InboundRequest request, InboundRequestInfo requestInfo)
     {
         Object[] params;
-
-        if( methodAccepts( method, request ) )
+        Method proxyMethod = requestInfo.getProxyMethod().orElseThrow(() -> new InboundRequestException("Missing proxy method, requestInfo: " + requestInfo));
```
cklellie, line 35: why not allow pass through of real method if no proxy?

casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/buffer/BufferHandler.java
```diff
@@ -30,12 +30,11 @@ public interface BufferHandler extends Prioritisable
...
      * Convert an inbound request to the appropriate {@link ServiceCallInfo}.
      *
-     * @param p the proxy for service invocation.
-     * @param method that the buffer will be dispatched to.
      * @param request that contains the buffer.
+     * @param requestInfo further request information
      * @return the transformed buffer and service call information.
      */
-    ServiceCallInfo fromRequest(Proxy p, Method method, InboundRequest request );
+    ServiceCallInfo fromRequest( InboundRequest request, InboundRequestInfo requestInfo );
```
cklellie: Ordering should be info first. just as it is was before, then also matches method below toResponse which takes info first.

casual/casual-inbound-handler-api/src/test/groovy/se/laz/casual/jca/inbound/handler/buffer/PassThroughBufferHandlerTest.groovy
```diff
@@ -18,10 +20,22 @@ class PassThroughBufferHandlerTest extends Specification
...
 
     def setup()
     {
         instance = new PassThroughBufferHandler()
+        proxyService = Mock( TestService )
+
+        Class[] c = new Class[1]
+        c[0] = TestService.class
+
+        jndiObject = (Proxy)Proxy.newProxyInstance(
```
cklellie: not needed - just replace with `Mock(Proxy)`, similarly the ForwardingInvocationHandler class is then not needed.

casual/casual-inbound-handler-fielded-buffer/src/main/java/se/laz/casual/jca/inbound/handler/buffer/fielded/FieldedBufferHandler.java
```diff
@@ -31,24 +32,26 @@ public boolean canHandleBuffer(String bufferType)
...
 
     @Override
-    public ServiceCallInfo fromRequest(Proxy p, Method method, InboundRequest request)
+    public ServiceCallInfo fromRequest(InboundRequest request, InboundRequestInfo requestInfo)
     {
         Object[] params;
-        if( methodAccepts( method, request ) )
+        Method proxyMethod = requestInfo.getProxyMethod().orElseThrow(() -> new InboundRequestException("Missing proxy method, requestInfo: " + requestInfo));
+        Method realMethod =  requestInfo.getRealMethod().orElseThrow(() -> new InboundRequestException("Missing real method, requestInfo: " + requestInfo));
+        if( methodAccepts( proxyMethod, request ) )
```
cklellie: If we have the real, why would we use the proxy in these accepts methods? Couldn't we just retrieve the real and pass that through, otherwise pass the proxy?

casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/buffer/PassThroughBufferHandler.java
aliothe: Not sure where that would happen?
For our PassthroughBufferHandler we expect the method invocation to be carried out using the proxy method no?

cklellie: true, leave it be.

build.gradle
```diff
@@ -36,13 +36,19 @@ allprojects{
...
-  }
+    }
 }
 
 subprojects{
   apply from: "$rootProject.projectDir/versions.gradle"
   apply plugin: 'java-library'
   apply from: "$rootProject.projectDir/publish.gradle"
+  test {
+	useJUnitPlatform()
```
TobiasLeo: Check the indents here. Looks like some tab characters have snuck in. Also some junk whitespace on line 39 ^

casual/casual-inbound-handler-api/src/test/java/se/laz/casual/jca/inbound/handler/test/ForwardingInvocationHandler.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2017 - 2018, The casual project. All rights reserved.
```
TobiasLeo: (c) 2017 - 2018 on a new file?

casual/casual-inbound-handler-api/src/test/groovy/se/laz/casual/jca/inbound/handler/buffer/InboundRequestInfoTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+import java.lang.reflect.Proxy
+
+class InboundRequestInfoTest extends Specification
+{
+   def 'creation'()
+   {
+      given:
+      def methodName = 'toString'
+      Method proxyMethod = String.class.getMethod(methodName)
+      Method realMethod = String.class.getMethod(methodName)
```
cklellie: Should be a different value.

casual/casual-inbound-handler-api/src/test/groovy/se/laz/casual/jca/inbound/handler/buffer/PassThroughBufferHandlerTest.groovy
```diff
@@ -18,10 +20,22 @@ class PassThroughBufferHandlerTest extends Specification
...
     @Shared Method echoString = TestService.getMethod( "echo", String.class )
+    @Shared TestService proxyService
+    @Shared Proxy jndiObject
 
     def setup()
     {
         instance = new PassThroughBufferHandler()
+        proxyService = Mock( TestService )
+
+        Class[] c = new Class[1]
```
cklellie: these two lines are now unused.

TobiasLeo (approved):

cklellie (approved):

PR #82, opened 2023-10-19
buggfix/2.2/pass real method to buffer handlers

aliothe:
The real method needs to be available for buffer handlers since annotations may only exist there and not on the proxy method.
We also future proof the API by using a wrapper class for the inbound request info that can be further expanded upon when need be.
    
    Other changes:
      * spock, same changes as for 3.2
      * using bytebuddy instead of cglib since it gives better error messages when things go wrong
      * up version of netty to 4.1.100

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,12 @@ subprojects{
   apply from: "$rootProject.projectDir/versions.gradle"
   apply plugin: 'java-library'
   apply from: "$rootProject.projectDir/publish.gradle"
+  test {
+    useJUnitPlatform()
+      testLogging {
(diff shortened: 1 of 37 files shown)


cklellie (commented):
Did you check it still works with java 11? Thinking the spock and cglib changes you made in the 3.2 you mentioned were for 17 mainly.

aliothe:
> Did you check it still works with java 11? Thinking the spock and cglib changes you made in the 3.2 you mentioned were for 17 mainly.

I am building it with java 8, that works fine.

cklellie (approved):

PR #83, opened 2023-10-30
Buggfix/2.2/honour tpnoreply

aliothe:
A non transactional tpacall with flags TPNOREPLY should never return nor expect an answer.


TobiasLeo (approved):
Looks good to me

PR #84, opened 2023-11-16
buggfix/3.2/honour tpnoreply (#83)

aliothe:
A non transactional tpacall with flags TPNOREPLY should never return nor expect an answer


TobiasLeo (approved):
Looks good to me

PR #89, opened 2024-01-30
feature/3.2/logging

aliothe:
* Log all available information in inbound
* Enable setting the netty log handler level for outbound/inbound and reverse inbound

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/api/util/PrettyPrinter.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/util/PrettyPrinter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, The casual project. All rights reserved.
+ * Copyright (c) 2021 - 2024, The casual project. All rights reserved.
  *
  * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
  */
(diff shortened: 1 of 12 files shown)


casual/casual-api/src/main/java/se/laz/casual/api/util/PrettyPrinter.java
```diff
@@ -9,21 +9,35 @@
...
-        return String.format("%s:%s:%s",toHex(xid.getGlobalTransactionId()), toHex(xid.getBranchQualifier()), xid.getFormatId());
+        return String.format("%s:%s:%s", toHex(xid.getGlobalTransactionId()), toHex(xid.getBranchQualifier()), xid.getFormatId());
     }
 
     public static String casualStringify(UUID uuid)
     {
         return String.format("%x%x", uuid.getMostSignificantBits(), uuid.getLeastSignificantBits());
     }
 
+    public static String print(UUID correlationId, UUID execution)
```
TobiasLeo: I'm not sure about the name print. It doesn't print anything. If anything I'd rather call it format since it outputs a formatted string

aliothe: sounds good to me, format that is

TobiasLeo (approved):
Looks good to me

PR #97, opened 2024-02-29
Feature/3.2/reverse inbound prepare commit rollback own threads

aliothe:
Reverse inbound is in fact a client and as such we do not want prepare/commit/rollback to run on any of nettys networking threads.
This is especially true since in fact an inbound prepare can result in an oubound prepare call via CasualXAResource, that then blocks that thread until an answer is received.

All of the operations are expected to be short running as they are not service calls, service calls stil run on the supplied work manager.


TobiasLeo (commented):
Looks good to me

TobiasLeo (approved):

Issue #44, opened 2022-10-26
Configuration with envs refactoring


cklellie:
With the addition of the ConfigurationService / Configuration option, we have started to consolidate how the casual jca is configured using files and or environment variables.
There are however a few places where direct environment variable  lookups are still performed. These should be refactored and added into the centrallise ConfigurationService.

cklellie:
Completed as part of version 3.2.47.

PR #99, opened 2024-03-13
feature/3.2/metric events

aliothe:
* Event Server - will only ever be available on 3.2
* Sonar version bump
* Gradle version bump
* System lambda version bump


event-server.md
```diff
@@ -0,0 +1,31 @@
+# Event server
+
+The purpose of the event server is to be able to access information regarding service calls as they happen - both inobund and outbound.
```
TobiasLeo: inobund->inbound

casual/casual-api/src/main/java/se/laz/casual/config/Configuration.java
```diff
@@ -45,6 +48,11 @@ public List<ReverseInbound> getReverseInbound()
         return null == reverseInbound ? Collections.emptyList() : Collections.unmodifiableList(reverseInbound);
     }
 
+    public Optional<EventServer> getEventServer()
+    {
+        return Optional.ofNullable(eventServer);
+    }
+
     @Override
     public boolean equals(Object o)
```
cklellie, line 57: Regenerate equals and hashcode to include event server.

casual/casual-api/src/main/java/se/laz/casual/config/EventServer.java
```diff
@@ -0,0 +1,73 @@
...
+    public int getPortNumber()
+    {
+        return portNumber;
+    }
+
+    public boolean isUseEpoll()
+    {
+        return useEpoll;
+    }
+
```
cklellie, line 30: missing equals and hashcode implementations.

casual/casual-api/src/test/groovy/se/laz/casual/config/CasualConfigTest.groovy
```diff
@@ -180,4 +181,25 @@ class CasualConfigTest extends Specification
...
+      String config = new File( "src/test/resources/" + file ).getText( StandardCharsets.UTF_8.name(  ) )
+      Configuration expected = Configuration.newBuilder()
+              .withEventServer(EventServer.createBuilder().withPortNumber(portNumber).withUseEpoll(useEpoll).build())
+              .build()
+
+      when:
+      Configuration actual = JsonProviderFactory.getJsonProvider(  ).fromJson( new StringReader( config ), Configuration.class )
+
+      then:
+      actual == expected
```
cklellie, line 197: interestingly, this is working because event part currently isn't with in the equals check.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package se.laz.casual.event;
+
+import se.laz.casual.spi.Prioritisable;
+
+public interface ServiceCallEventHandler extends Prioritisable
+{
+    /**
+     * Non blocking operation
+     * @param event
```
cklellie: not a complete javadoc - need description for event param.
Equally, if you're going to have javadoc, add a bit more of a description to what this method does.
Though perhaps changing the naming would make it more obvious.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param serviceName The service name
+     * @param code The state for the service call
+     * @param pendingMicroseconds The pending time in microseconds
+     * @param start The start time
+     * @param end The end time
+     * @param order The order
+     */
+    public  void createAndPostEvent(Xid xid, UUID execution, String parentName, String serviceName, ErrorState code, long pendingMicroseconds, Instant start, Instant end, Order order)
+    {
+        Objects.requireNonNull(xid, "xid can not be null");
```
cklellie: Would it not be nicer to have this validation inside the `builder.build()` method?
Equally, whats the point of having a builder if you then expose this method with 9 parameters? 
Shouldn't the callee build the object and then "publish" it?

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventHandler.java
```diff
@@ -0,0 +1,23 @@
...
+     */
+    void put(ServiceCallEvent event);
```
cklellie: Naming - Publish? Fire? Emit?

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Blocking operation
+     * @return A service call event
+     */
+    ServiceCallEvent take();
```
cklellie: Naming - consume? receive?

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventHandlerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Retrieve the most appropriate {@link ServiceCallEventHandler} based on it's {@link Priority}.
+     *
+     * If there is no registered handler a {@link CasualRuntimeException} is thrown.
+     *
+     * @return the appropriate handler.
+     */
+    public static ServiceCallEventHandler getHandler()
+    {
+        return getHandlers().stream()
```
cklellie, line 30: Currently there can be multiple, hence needing a static LinkedQueue. Is that a constraint we want to enforce upon all other implementation in the future?
Or should this be cached after first call?

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
```diff
@@ -0,0 +1,113 @@
...
+                                               .withService(serviceName)
+                                               .withCode(code)
+                                               .withStart(ChronoUnit.MICROS.between(Instant.EPOCH, start))
+                                               .withEnd(ChronoUnit.MICROS.between(Instant.EPOCH, end))
+                                               .withOrder(order)
+                                               .withPending(pendingMicroseconds)
+                                               .withParent(parentName)
+                                               .build());
+        }
+        catch(Exception e)
```
cklellie: From where exactly can this exception occur? The `between` in the Chrono? How? Feels like this would be an all or nothing situation for every event call, so perhaps we should guard against it on startup of the event server, rather than flooding logs on calls and needing this catch?

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
```diff
@@ -0,0 +1,113 @@
...
+                serviceName, parentName, PrettyPrinter.casualStringify(xid), PrettyPrinter.casualStringify(execution), code.name(), start, end, order.getValue());
+    }
+
+    private void post(ServiceCallEvent event)
+    {
+        try
+        {
+            handler.put(event);
+        }
+        catch(NoServiceCallEventHandlerFoundException e)
```
cklellie: How does this exception occur? The handler is injected into this component via the `of` constructor.

casual/casual-event-api/src/main/java/se/laz/casual/event/Process.java
```diff
@@ -0,0 +1,30 @@
...
+    private static final Logger LOG = Logger.getLogger(Process.class.getName());
+    private static final long CAN_NOT_GET_PID = -1L;
+    private static long currentPID = CAN_NOT_GET_PID;
+    private Process()
+    {}
+    public static long pid()
+    {
+        try
+        {
+            currentPID = currentPID == CAN_NOT_GET_PID ? ProcessHandle.current().pid() : currentPID;
```
cklellie, line 22: I'm lost, why are we re assigning the pid with every operation? Is it not more efficient to just guard with an if and return otherwise calculate? I don't think JIT will be able to optimise away this.
I also don't see it being possible for it to fail and then work later. So a singleton in which pid is calculate on first request would perhaps be a better pattern.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.transaction.xa.Xid;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class ServiceCallEventPublisher
```
cklellie, line 14: I am lost by the value of this additional layer of abstraction. Are you just restricting access to only publish? 

Also the ServiceCallHandler, isn't really a Handler, its more a store. Perhaps that the distinction. That handler is a store, and this is a publication interface to that store? Then we have the event server which is the consumer?

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServerConnectionInformation.java
```diff
@@ -0,0 +1,106 @@
...
+
+        public Builder withServerInitialization(ServerInitialization serverInitialization)
+        {
+            this.serverInitialization = serverInitialization;
+            return this;
+        }
+
+        public EventServerConnectionInformation build()
+        {
+            logHandlerEnabled = Boolean.parseBoolean(System.getenv(USE_LOG_HANDLER_ENV_NAME));
```
cklellie, line 101: one to add to the list in https://github.com/casualcore/casual-java/issues/44 i guess.

casual/casual-event-server/src/main/java/se/laz/casual/event/server/messages/LogonRequest.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+package se.laz.casual.event.server.messages;
+
+import se.laz.casual.api.CasualTypeException;
+
+import java.util.Arrays;
+
+public enum LogonRequest
```
cklellie: Naming here of "logon" for me implies credentials passing. 
As you described it before, it's just an initialisation handshake, very remeniscent of SMTP `EHLO` / `HELO`.
Perhaps `InitRequest`, `ConnectRequest`, `ReceiveRequest`?

casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
```diff
@@ -94,7 +99,10 @@ private void issueCallNoReply()
     {
         try
         {
+            Instant start = Instant.now();
             callService();
+            Instant end = Instant.now();
+            getEventPublisher().createAndPostEvent( message.getXid(), message.getExecution(), message.getParentName(), message.getServiceName(), ErrorState.OK, startupTimeFuture.join(), start, end, Order.CONCURRENT);
```
cklellie: From previous discussions with Fredrik, Inbound should be `sequential` and outbound should be `concurrent`, is that not the case anymore?

casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
```diff
@@ -94,7 +99,10 @@ private void issueCallNoReply()
     {
         try
         {
+            Instant start = Instant.now();
```
cklellie: Do we have to do this everywhere - couldn't we utilise ServiceCallWorkListener to do this less intrusively?
As we receive in the `WorkEvent`, the `Work` object, so should be able to get everything out from there.

cklellie: Could we just use `System.nanoTime` and truncate to micro, removing the Chrono RuntimeException issue I mentioned earlier?
In fact if we use the builder more - you could just call a no arg `start` and then after call a no arg `end` and hide that calculation completely.

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
```diff
@@ -128,14 +130,16 @@ public void serviceCallRequest(CasualNWMessage<CasualServiceCallRequestMessage>
...
         }
         boolean isTpNoReply = message.getMessage().getXatmiFlags().isSet(AtmiFlags.TPNOREPLY);
-        CasualServiceCallWork work = new CasualServiceCallWork(message.getCorrelationId(), message.getMessage() , isTpNoReply);
+        CompletableFuture<Long> startupTimeFuture = new CompletableFuture<>();
+        CasualServiceCallWork work = new CasualServiceCallWork(message.getCorrelationId(), message.getMessage() , isTpNoReply, startupTimeFuture);
 
         try
         {
-            long startup = !isTpNoReply && isServiceCallTransactional( xid ) ?
+            long startupInMilliseconds = !isTpNoReply && isServiceCallTransactional( xid ) ?
```
cklellie: `startWork` return can be -1 according to the javadoc, can we accept that, or rather calculate ourselves using the Listener as previously mentioned using `workAccepted` and `workStarted` options.

casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
```diff
@@ -108,18 +116,24 @@ private void issueCall()
...
             replyBuilder
                     .setError(reply.getErrorState())
                     .setTransactionState(reply.getTransactionState())
                     .setUserSuppliedError( reply.getUserSuppliedErrorCode() );
+            getEventPublisher().createAndPostEvent( message.getXid(), message.getExecution(), message.getParentName(), message.getServiceName(), reply.getErrorState(), startupTimeFuture.join(), start, end, Order.CONCURRENT);
         }
         catch( ServiceHandlerNotFoundException e )
         {
+            Instant end = Instant.now();
+            getEventPublisher().createAndPostEvent( message.getXid(), message.getExecution(), message.getParentName(), message.getServiceName(), ErrorState.TPENOENT, startupTimeFuture.join(), start, end, Order.CONCURRENT);
```
cklellie: This all feels very intrusive and repetative. I really do think the work should be completed prior to publishing any form of event, otherwise there will be an overhead, even if relatively small.

casual/casual-jca/build.gradle
```diff
@@ -39,13 +39,15 @@ dependencies {
...
+    implementation project(':casual:casual-event-api')
+    implementation project(':casual:casual-event-server')
     implementation libs.netty
     compileOnly libs.javaee_api
     compileOnly libs.gson
 
     resourceAdapter( project(':casual:casual-network') )
     resourceAdapter project(':casual:casual-network-protocol')
-    resourceAdapter project(':casual:casual-inbound-api')
+    resourceAdapter (project(':casual:casual-event-server')){ transitive = false }
```
cklellie, line 50: What happened here? Why are things transitive and then removed `casual-inbound-api` ?

deployment-wildfly.md
```diff
@@ -1,15 +1,15 @@
 # Wildfly Deployment
 
 ## Wildfly version
-Note that these examples are tested with wildfly `13.0.0.Final`.
+Note that these examples are tested with wildfly `CASUAL_VERSION` `31.0.1.Final`.
```
cklellie, line 4: ? what does `CASUAL_VERSION` mean here?

deployment-wildfly.md
```diff
@@ -68,18 +68,20 @@ Run the following command via jboss-cli:
...
-    jndi-name=eis/casualConnectionFactoryTjohej,\
-    min-pool-size=5, initial-pool-size=5, max-pool-size=5,\
-    #interleaving=true,\
+    jndi-name=eis/casualConnectionFactory,\
+    min-pool-size=100, initial-pool-size=100, max-pool-size=100,\
     enabled=true)
 
-$connectionDefinitionNode/config-properties=HostName:add(value=192.168.99.100)
-$connectionDefinitionNode/config-properties=PortNumber:add(value=7771)
+$connectionDefinitionNode/config-properties=HostName:add(value=some-host-that-speaks-casual)
```
cklellie: I think concrete examples here were better

casual/casual-jca/src/main/java/se/laz/casual/jca/service/CasualServiceCaller.java
```diff
@@ -73,23 +79,38 @@ public ServiceReturn<CasualBuffer> tpcall(String serviceName, CasualBuffer data,
...
+                casualNWMessageCompletableFuture.whenComplete((v, e) -> {
+                            if (null != e)
+                            {
+                                LOG.finest(() -> "service call request failed for corrid: " + PrettyPrinter.casualStringify(corrId) + SERVICE_NAME_LITERAL + serviceName);
+                                f.completeExceptionally(e);
+                                return;
+                            }
+                            LOG.finest(() -> "service call request ok for corrid: " + PrettyPrinter.casualStringify(corrId) + SERVICE_NAME_LITERAL + serviceName);
+                            final Instant end = Instant.now();
+                            getEventPublisher().createAndPostEvent( xid, execution, "", serviceName, v.getMessage().getError(), 0, start, end, Order.SEQUENTIAL);
```
cklellie: Outbound should be `C`.

cklellie: What about reverse inbound - what should that be...

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServer.java
```diff
@@ -0,0 +1,68 @@
...
+        this.channel = channel;
+    }
+
+    public static EventServer of(EventServerConnectionInformation connectionInformation)
+    {
+        Objects.requireNonNull(connectionInformation, "connectionInformation can not be null");
+        ChannelGroup connectedClients = new DefaultChannelGroup(GlobalEventExecutor.INSTANCE);
+        Channel ch =  connectionInformation.getServerInitialization().init(connectionInformation, connectedClients);
+        final ServiceCallEventHandler serviceCallEventHandler = ServiceCallEventHandlerFactory.getHandler();
+        MessageLoop messageLoop = DefaultMessageLoop.of(connectedClients, serviceCallEventHandler::take);
```
cklellie: We have spoken before about this, you have only one implementation, why are we using a supplier, just inject the interface/component, not the function. It just adds to the cognitive load of abstractions trying to follow the code. In this specific case you also hide the blocking nature of the `get` which calls a `take`. It's confusing.

deployment-wildfly.md
aliothe: CASUAL_VERSION is a symbol that is set and used later on down below in cli script snippets.
So next time we can just adjust that instead of having the value in each place.

casual/casual-event-server/src/test/groovy/se/laz/casual/event/server/handlers/FromJSONLogonDecoderTest.groovy
```diff
@@ -0,0 +1,62 @@
...
+import io.netty.buffer.ByteBuf
+import io.netty.channel.Channel
+import io.netty.channel.ChannelHandlerContext
+import io.netty.channel.group.ChannelGroup
+import io.netty.util.CharsetUtil
+import se.laz.casual.api.CasualTypeException
+import se.laz.casual.event.server.messages.ConnectRequestMessage
+import spock.lang.Specification
+
+class FromJSONLogonDecoderTest extends Specification
```
cklellie: Test class name and test names.

cklellie (approved):

Issue #96, opened 2024-02-23
Handle exception in inbound work listener


aliothe:
https://github.com/casualcore/casual-java/blob/2.2/casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/ServiceCallWorkListener.java

In workCompleted, it could be exceptional and then work.getResponse is null.

aliothe:
Also handle if there is no actual response so we do not try to write any null values

cklellie:
Page 235 of the connection 1.7 spec shows a nice state diagram of what is expected to happen after work is submitted.

There are multiple parts to this issue.

1. ServiceHandler implementations should "never" allow an exception to bubble out of the workmanager.

This should be checked to ensure all code paths are within try blocks. CasualServiceHandler currently catches Throwable, so it should not break out, but there are a few calls prior to that which might be missed.
All exceptions should be handled within the ServiceHandler and return appropriate responses, without allowing the exception to bubble up.
 
2. WorkCompletedException should be checked for inside workCompleted in case there is something exceptional.

In the rare event that a work does result in an exception - could be due to faulty implementations of ServiceHandler or ServiceHandlerExtention for example. The WorkEvent exception should be checked and handled similarly to within the ServiceHandler i.e. WARN and return empty buffer with TPESVCERR.

3. workRejected should be implemented incase this occurs.

Currently this is a noop thought it should be implemented and handled similarly to within ServiceHandler i.e. WARN and return empty buffer with TPESVCERR.

PR #100, opened 2024-03-15
work rework, do not block service calls

aliothe:
Note, this is another branch that could go into the metric events branch if the changes are sufficient.
Created a new branch for these changes since they are rather large so that we can discuss it here more than anything else.

Things to ponder:
Why do we not use the async version since the only benefit now is a ms time that we get back, which can be -1, and we log that one.
Otherwise, we just block the calling thread for no reason.

Re pending:
https://docs.oracle.com/javaee/7/api/javax/resource/spi/work/WorkListener.html#workAccepted-javax.resource.spi.work.WorkEvent-
void workAccepted(WorkEvent e)
Invoked when a Work instance has been accepted.
Ok, so not useful
void workStarted(WorkEvent e)
Invoked when a Work instance has started execution. This only means that a thread has been allocated.
Also means that it's not exactly before the actual service call.
We do however have that point in the actual work, right before we make the call
Then it's when we should have the end time, afer the call or in WorkCompleted ( which will be some time later on)
Two things there, pending and start/end time for the actual call - if that also should be removed from the work task itself

The thread that gets block currently, both for normal inbound and reverse for a service call, is a netty thread.
I can't see a reason why not use the async version?

Then for inbound we really need to fix the exception handling black hole and also take care if there is no response when we expect a response:
https://github.com/casualcore/casual-java/issues/96 - however, that should be in a separate feature since it also needs to go into 2.2.

Another thing, there is no issue for this yet but - in case we get a WorkException when dispatching the actual work - as far as I can see, that is not handled at all?

### DIFF
--- a/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
+++ b/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
@@ -12,10 +12,6 @@ import se.laz.casual.api.buffer.type.ServiceBuffer;
 import se.laz.casual.api.flags.ErrorState;
 import se.laz.casual.api.flags.TransactionState;
 import se.laz.casual.api.network.protocol.messages.CasualNWMessage;
-import se.laz.casual.event.Order;
-import se.laz.casual.event.ServiceCallEvent;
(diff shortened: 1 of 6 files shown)


cklellie (approved):
Looks good to merge back into the other PR.

PR #102, opened 2024-03-19
Feature/3.2/metrics events/builder timer

cklellie:
Changed ServiceCallEvent builder to implement timer functionality to remove need for working with Instant objects inside the calling code.
Simplified usage of builder in work listener and removed duplication in caller.
Added fix for issue with module access on windows for using system lambda.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,7 +60,7 @@ subprojects{
   }
 
   test{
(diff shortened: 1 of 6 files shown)


aliothe (commented):
Looks fine

aliothe (approved):

PR #103, opened 2024-03-21
Feature/3.2/sphinx docs

aliothe:
sphinx docs

Javadoc is linked as is, had a stab with doxygen and other tools first but it's just painful.


docs/source/index-sphinx.rst
```diff
@@ -0,0 +1,13 @@
...
+   contain the root `toctree` directive.
+
+casual-java documentation
+=======================================
+
+.. toctree::
+   :titlesonly:
+   
+   Casual Java Integration <casual-docs/README>
+   javadoc <index>
```
TobiasLeo: Would probably look better as `Javadoc` with a capital `J`

docs/source/conf.py
```diff
@@ -0,0 +1,39 @@
...
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+import sphinx_rtd_theme
+
+project = 'casual-java'
+copyright = '2024, casualproject'
+author = 'casualproject'
+release = '3.2.33'
```
TobiasLeo: Would it be possible to derive the docs version from the casual-java version? Will they always  be the exact same?

aliothe: Yeah, we can get the current version from versions.gradle

TobiasLeo (approved):

PR #104, opened 2024-04-10
Feature/2.2/reverse inbound connection attempt backoff

TobiasLeo:


casual/casual-api/src/main/java/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWork.java
```diff
@@ -20,20 +23,26 @@ public class RepeatUntilSuccessTaskWork<T> implements Work
     private final Supplier<T> supplier;
     private final Consumer<T> consumer;
     private final Supplier<WorkManager> workManagerSupplier;
+    private int failures = 0;
+    private final long maxBackoffMillis;
+    ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
```
cklellie: Can you instead utilise the inbuilt `ManagedScheduleExecutorService`? Or does this not work in both weblogic and wildfly?

casual/casual-network/src/main/java/se/laz/casual/network/inbound/reverse/ReverseInboundConnectionInformation.java
```diff
@@ -33,6 +33,7 @@ public class ReverseInboundConnectionInformation
     private final String domainName;
     private final Class<? extends Channel> channelClass;
     private boolean useEpoll;
+    private final long maxConnectionBackoffMillis;
```
cklellie: Now we are in the context of a `ReverseInboundConnectionInformation`, do we need to repeat the `Connection` in the variable? `maxBackoffMillis` for readability/brevity perhaps?

reverse-inbound.md
```diff
@@ -23,3 +28,7 @@ host:port. If no size is specified then only one such connection is created per
 
 After the connection has succeeded, this connection now works as it was a normal *inbound* connection.
 
+If `maxConnectionBackoffMillis` is specified for a reverse inbound connection then that value will be the maximum
+connection backoff time if the target host isn't available. When nothing is configured a default value of 30 seconds for
+maximum backoff time will be used. The backoff is exponential and will start att 1/10 of the maximum value and increase
```
cklellie: It is not exponential, it is linear, 30s would be 3s, 6s, 9s,12s... with a simple exponential backoff you use a fixed factor of 2 to multiple some amount for each retry up until a limit, say 3s, 6s, 12s, 24s... 
Which do we want to implement here?

casual/casual-api/src/test/groovy/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWorkTest.groovy
```diff
@@ -51,7 +51,7 @@ class RepeatUntilSuccessTaskWorkTest extends Specification
          workManager
       }
       when:
-      RepeatUntilSuccessTaskWork taskWork = RepeatUntilSuccessTaskWork.of(supplier, consumer, workManagerSupplier)
+      RepeatUntilSuccessTaskWork taskWork = RepeatUntilSuccessTaskWork.of(supplier, consumer, workManagerSupplier, 1000)
```
cklellie: I think it would be nice to have an additional assertions and/or test cases here to see that the backoff is actually working.
Potentially with different values for the backoff.

casual/casual-api/src/main/java/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWork.java
```diff
@@ -90,4 +102,20 @@ private void scheduleWork()
             }
         }
     }
+
+    private long getBackoffMillis()
```
aliothe: The backoff logic itself could be refactored into its own logical unit so that we can have unit tests that tests that functionality by itself.

reverse-inbound.md
TobiasLeo: The implementation isn't linear, it goes from 1/10th of the max, then 1/9th, 1/8th ... until it reaches the max with 1/1th of the max which is the max value. Basically the simplest exponential-like algorithm I could cook up that could deterministically reach the maxBackoff in n steps where n was arbitrarily chosen as 10. The documentation just isn't clear and maybe should be more precise if the current backoff algorithm is used. The actual code probably has a better comment describing how the backoff works

casual/casual-network/src/main/java/se/laz/casual/network/inbound/reverse/ReverseInboundConnectionInformation.java
TobiasLeo: Yeah, your are probably right. That's what I had at first and felt like it might be unclear, but it really shouldn't be in the context of the class it's in. I'll fix it

casual/casual-api/src/main/java/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWork.java
TobiasLeo: I remember having problems looking up the `ManagedScheduledExecutorService` from jndi on WildFly and can't just `@Resource` it in in this context so I used this instead to achieve something that works. I can have another quick look at why it failed and if I can manage to use `ManagedScheduledExecutorService` in a portable manner

casual/casual-api/src/main/java/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWork.java
TobiasLeo: Absolutely, I'll rework it into some `Backoff` class of some sort, maybe it can even encapsulate the counted failures

aliothe: It can indeed, sounds good!

casual/casual-api/src/main/java/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWork.java
aliothe: wls has extensions to the JCA specification - one of them being that CDI works.
It is not in the specification and thus will never work on wildfly unless they also implement an extension which I doubt that they will.

reverse-inbound.md
cklellie: ah, i read the comment on line 110 `// Backoff in steps of 1/10, 2/10 .. 8/10, 9/10 until maxBackoffMillis is reached`
which is linear, and missed the actual implementation was different. :) If you update the comment and change the documentation here accordingly as either way it's not really exponential.

TobiasLeo: I actually though that comment was correct, but it really isn't. Realized it while reworking the code now as well. I'll update the documentation and code-comment to be correct to not call it exponential since it really isn't and then describe what the backoff-algorithm really does

casual/casual-api/src/test/groovy/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWorkTest.groovy
TobiasLeo: I'll add that. Probably fits best into its own separate test case since it's out of the scope of the current tests that just needed the argument because he constructing method was expanded

casual/casual-api/src/main/java/se/laz/casual/api/util/work/RepeatUntilSuccessTaskWork.java
TobiasLeo: After looking at it some more: on WildFly I can't just lookup `java:comp/DefaultManagedScheduledExecutorService`, maybe because I'm not in an EJB context currently or something, I don't know. What does yield a result is to lookup the jboss default `java:jboss/ee/concurrency/scheduler/default` which gives me a `ManagedScheduledExecutorService` that works. Seems we already have some sort of adjacent hack around getting our hands a `ManagedExecutorService` via https://github.com/casualcore/casual-java/blob/f913f5c0dccd5fea47bfaad1537c02675636353a/casual/casual-network/src/main/java/se/laz/casual/network/outbound/JEEConcurrencyFactory.java#L28

Would it really be worth it to extend the hack to be able to fetch `ManagedScheduledExecutorService` instances as well?

aliothe: The problem being that we do not know the direct names for WLS - thus that only works for wildfly.
I remember that the these direct names are actually documented by wildfly, they are not by wls.

Which is pretty much the comment in that file I see :)

aliothe: Since we now have a split between 3.2 and 2.2 - we could say that 2.2 is wls only going forward.
That would allow us to use CDI stuff on 2.2 and the direct names on 3.2.
If we want to do that, @cklellie @TobiasLeo ?

casual/casual-network/src/main/java/se/laz/casual/network/outbound/JEEConcurrencyFactory.java
```diff
@@ -62,4 +70,48 @@ public static ManagedExecutorService getManagedExecutorService()
...
+                return getSharedJavaUtilScheduledExecutor();
+            }
+        }
+    }
+
+    private static synchronized ScheduledExecutorService getSharedJavaUtilScheduledExecutor()
+    {
+        if (null == sharedScheduledExecutor)
+        {
+            sharedScheduledExecutor = Executors.newScheduledThreadPool(10);
```
cklellie: 10, because? Just curious about sizing what this impacts, would it need to be higher than the number of reverse connections or? Which then leads me to wonder, what is the configuration of the weblogic / wildfly versions, do they need to be configured similarly. Just would be good to know from a documentation perspective, in which case we might want to make this 10 configurable too.

TobiasLeo: Basically arbitrarily chosen. I felt like it needed to be able to handle a bunch of concurrent calls since it's a shared scheduler, but wanted to be able to push something before the weekend. I agree it should be configurable, but does it need to be added right now? This particular scheduler won't be used in a production case at the moment. If it isn't made configurable now an issue should be created to add config later, though

cklellie: If you add quick integer env lookup, then we can fix it properly in the other issue related to getting everything inside ConfigurationService: https://github.com/casualcore/casual-java/issues/44

casual/casual-network/src/main/java/se/laz/casual/network/outbound/JEEConcurrencyFactory.java
```diff
@@ -62,4 +70,48 @@ public static ManagedExecutorService getManagedExecutorService()
...
+     * it will fetch the default through java:comp/DefaultManagedScheduledExecutorService
+     * which normally shouldn't be accessible in this context.
+     * <p>
+     * If both of those fail a ScheduledExecutorService instance will be created
+     * through the standard java.util.concurrent.Executor which will share the
+     * same ScheduledExecutorService among all users.
+     *
+     * @return A ScheduledExecutorService instance.
+     */
+    public static ScheduledExecutorService getManagedScheduledExecutorService()
```
aliothe, line 90: getManagedScheduledExecutorService should perhaps be renamed since it can in fact return a plain vanilla ScheduledExecutorService?
Would geScheduledExecutorService suffice - the comment on the API tells the rest of the story?

cklellie: Equally, you could argue that in that unlikely event, we are "managing" access to it, as it will be reusable, so the naming is fine and the existing comment explains the situation well.

aliothe: Managed in the context is with regards to JEE though.
It's a minor thing though, my context is when using our stuff standalone outside of any appserver  ( Quarkus etc).

TobiasLeo: I'll leave it as is for now. It's not an external api and can easily be renamed in the future if we find a better name that is both descriptive and doesn't infringe on JEE nomenclature

aliothe (approved):

PR #105, opened 2024-04-12
Feature/3.2/event service testability

cklellie:
Changes for testability of EventServer.

- Added `withStart` and `withEnd` to generate fake events.
- Moved time caluations into existing InstantUtil.
- Added configuration for EventServer shutdown, so default of 2 second quietPeriod is not longer mandatory when cleaning up during unit testing.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/config/Configuration.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/config/Configuration.java
@@ -78,10 +78,11 @@ public class Configuration
     public String toString()
     {
         return "Configuration{" +
-                "inbound=" + getInbound() +
-                ", domain=" + getDomain() +
-                ", outbound=" + getOutbound() +
-                ", reverseInbound=" + getReverseInbound() +
+                "inbound=" + inbound +
+                ", domain=" + domain +
+                ", outbound=" + outbound +
+                ", reverseInbound=" + reverseInbound +
+                ", eventServer=" + eventServer +
                 '}';
     }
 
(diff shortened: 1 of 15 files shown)


casual/casual-api/src/main/java/se/laz/casual/config/Configuration.java
aliothe, line 1: Why this change?

casual/casual-api/src/main/java/se/laz/casual/config/Shutdown.java
```diff
@@ -0,0 +1,114 @@
+/*
+ * Copyright (c) 2021, The casual project. All rights reserved.
```
aliothe: 2024 no? :)

aliothe (approved):
Looks good besides 2021 instead of 2024 and I am curious as to why the toString method in Configuration changed to using the attributes instead of the methods?

casual/casual-api/src/main/java/se/laz/casual/config/Configuration.java
cklellie: I regenerated it to include event server, i missed it used methods not attributes. I'll change back to methods again.

casual/casual-event-server/src/test/groovy/se/laz/casual/event/server/EventServerConnectionInformationTest.groovy
TobiasLeo, line 1: License header?

PR #107, opened 2024-04-16
Feature/3.2/event client

aliothe:
Event client to consume events from the event server in an efficient and easy fashion.


event-client.md
```diff
@@ -0,0 +1,19 @@
...
+
+Example usage:
+
+```java
+EventClientInformation clientInformation = EventClientInformation.createBuilder()
+                                                                 .withChannelClass(EpollSocketChannel.class)
+                                                                 .withEventLoopGroup(new EpollEventLoopGroup())
+                                                                 .withConnectionInformation(new ConnectionInformation("localhost", 7289))
+                                                                 .build();
+EventClient client = EventClient.of(clientInformation, eventObserver, connectionObserver,  enableLogging);
```
cklellie: This is a far bit of boiler plate to just make a client. 
I would rather have something much simpler e.g.: `EventClient.newBuilder().address( "localhost").port(7774).observer( myobserver ).build()`.
Then the builder can take care of the rest by making everything default.
If you want to mess around with channel class and event loop groups and stuff then fine, but if you just want to connect, that's all i want.

aliothe: You do want to be able to choose which event loop and channel class you want to use, for several reasons.
1) They have different characteristics
2) Not all are supported on all platforms, only linux supports all currently

aliothe: Adding a builder to the client seems rather horrible honestly.

PR #108, opened 2024-04-17
Feature/3.2/event client

aliothe:
* Event client, simple and efficient client to handle service call events.
* Now also with a connect future for usage in test clients.


casual/casual-event-client/src/main/java/se/laz/casual/event/client/EventClientBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public final class EventClientBuilder
+{
+    private EventClientBuilder()
+    {}
+    public static Builder createBuilder()
+    {
+        return new Builder();
+    }
+    public static final class Builder
```
TobiasLeo, line 27: Do we really need an inner `Builder` class when it's already a separate `EventClientBuilder` class for the builder?

casual/casual-event-server/src/test/groovy/se/laz/casual/event/server/handlers/ExceptionHandlerTest.groovy
```diff
@@ -27,6 +30,7 @@ class ExceptionHandlerTest extends Specification
       ExceptionHandler exceptionHandler = ExceptionHandler.of(clients)
       when:
       exceptionHandler.exceptionCaught(ctx, new CasualRuntimeException("Ooopsie"))
+      println("${JsonProviderFactory.getJsonProvider().toJson(ConnectReplyMessage.of(ConnectReply.CONNECT_REPLY))}")
```
TobiasLeo: Does this `println` line serve a purpose for the test?

casual/casual-event-client/src/main/java/se/laz/casual/event/client/EventClientBuilder.java
aliothe: Strictly speaking no, it's more that it follows our usual builder pattern though it is a rather special builder.

cklellie: All classes with builders we currently have the builder as an inner. So if you want to follow our usual builder pattern that is where this should live, inside EventClient. The war over `createBuilder` vs `newBuilder` is a fight for another day though!

aliothe: I really do not want to stuff a 92 LOC builder into a 110 LOC EventClient - that will make that code so much more unreadable.
As you can use the EventClient perfectly fine without the builder - that's another reason to not stuff it into that class.

TobiasLeo: You could argue either way about the separated builder. While it does follow the same patterns as other builders by being, I feel like it's sort of weird and gives the impression we have a builder for a builder with the whole `EventClientBuilder.Builder` class

In the end I don't have a too strong opinion on the matter and how this builder is constructed could be done in multiple ways. Since we can still get a builder instance through the `static EventClientBuilder.Builder EventClient::createBuilder()` there shouldn't be much confusion regarding its use anyway

TobiasLeo (approved):

PR #112, opened 2024-04-26
allow real, optional, method as well

aliothe:
There is a need to be able to access the real method as well when using fielded and primitive return values.


cklellie (approved):
This will go back into 2.2 branch as well?

aliothe:
> This will go back into 2.2 branch as well?

Yes, it will go into 2.2 as well.

PR #114, opened 2024-05-07
Feature/3.2/casual test

cklellie:
Added casual-test module. Initial implementation of a casual embedded server for use in unit testing casual java.
Currently focused upon EventServer, though with abition for future iteration to allow developers to deploy inbound services to test them quickly during unit testing rather than waiting for integration testing.

Few fixes to make things work on windows by default with netty when running gradle builds.


casual-test.md
```diff
@@ -0,0 +1,16 @@
...
+
+The tools provided will as much as possible utilise existing code and components of the casual-java, though provide 
+additional test apis to allow for example, test data and fault injection.
+
+It must be light-weight and fast such that initialisation can be perform often within for example unit tests without
+significant overhead.
+
+It must be thread safe and allow multiple concurrent instances to run within the same jvm allowing concurrent/parallel
+runs of test which utilise seperate instances of the embedded server.
+
```
aliothe, line 16: Do we run tests in parallell today?

casual/casual-event-api/src/main/java/se/laz/casual/event/CasualServiceCallEventStore.java
```diff
@@ -15,7 +15,7 @@
  */
 public class CasualServiceCallEventStore implements ServiceCallEventStore
 {
-    private static final BlockingDeque<ServiceCallEvent> EVENTS = new LinkedBlockingDeque<>();
+    private final BlockingDeque<ServiceCallEvent> EVENTS = new LinkedBlockingDeque<>();
```
aliothe, line 18: The design of the actual implementation is fine, it should be static but now it is not static due to a test tool?
I do not like that at all.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventStoreFactory.java
```diff
@@ -9,30 +9,34 @@
...
      *
      * If there is no registered handler a {@link CasualRuntimeException} is thrown.
      *
+     * The store is cached based on the provided domainId.
+     *
+     * @param domainId unique id of the domain to which the ServiceCallEventStore belongs.
      * @return the appropriate handler.
      */
-    public static ServiceCallEventStore getHandler()
+    public static ServiceCallEventStore getStore(UUID domainId)
```
aliothe, line 32: Again, this changes the actual design of the event server.

aliothe: The domain in production code, in runtime, can not be several - it is always one.

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServerConnectionInformation.java
```diff
@@ -126,7 +126,7 @@ public ServerInitialization getServerInitialization()
     public static final class Builder
     {
         private int port;
-        private boolean logHandlerEnabled;
+        private Boolean logHandlerEnabled;
```
aliothe, line 129: why, it is either true or false - not maybe?

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServerConnectionInformation.java
```diff
@@ -170,9 +170,18 @@ public Builder withShutdownTimeout( long timeout )
...
+        public Builder withLogHandlerEnabled( Boolean enabled )
+        {
+            this.logHandlerEnabled = enabled;
+            return this;
+        }
+
         public EventServerConnectionInformation build()
         {
-            logHandlerEnabled = Boolean.parseBoolean(System.getenv(USE_LOG_HANDLER_ENV_NAME));
+            if( logHandlerEnabled == null )
```
aliothe, line 181: why?

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
```diff
@@ -1,10 +1,17 @@
 /*
- * Copyright (c) 2017 - 2018, The casual project. All rights reserved.
+ * Copyright (c) 2017 - 2024, The casual project. All rights reserved.
  *
  * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
  */
 package se.laz.casual.jca;
 
+import jakarta.resource.ResourceException;
+import jakarta.resource.spi.*;
```
aliothe: Please have all the imports, probably something wrong with your IntelliJ setting?

casual/casual-jca/src/main/java/se/laz/casual/jca/service/CasualServiceCaller.java
```diff
@@ -35,13 +35,8 @@
...
 import javax.transaction.xa.Xid;
 import java.time.Duration;
-import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
aliothe: Same thing here, we want all imports visible

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
```diff
@@ -0,0 +1,217 @@
...
+     * NB - can only be called when the server is not running.
+     */
+    public void start()
+    {
+        if( running.compareAndSet( false, true ) )
+        {
+            initialiseEventServer();
+            return;
+        }
+        throw new IllegalStateException( "Server is already started." );
```
aliothe, line 104: would it not be ok for start to be idempotent?

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    private void initialiseEventServer()
+    {
+        if( !this.eventServerEnabled )
```
aliothe, line 109: This can not really happen due to the CAS in start no, if it is only called from start that is?

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
```diff
@@ -0,0 +1,217 @@
...
+    /**
+     * Simulate a service call event being published.
+     *
+     * Can be used to test integration with event server.
+     *
+     * @param serviceCallEvent to be published.
+     */
+    public void publishEvent(ServiceCallEvent serviceCallEvent)
+    {
+        ServiceCallEventStore store = ServiceCallEventStoreFactory.getStore(this.domainId);
```
aliothe: Contract to make sure that the event is not in fact null?

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
```diff
@@ -0,0 +1,217 @@
...
+        return "CasualEmbeddedServer{" +
+                "eventServerPort=" + eventServerPort +
+                ", eventServerEnabled=" + eventServerEnabled +
+                ", running=" + running +
+                '}';
+    }
+
+    public static Builder newBuilder(CasualEmbeddedServer src )
+    {
+        return new Builder().eventServerPort( src.eventServerPort );
```
aliothe: Law of Demeter, if the only thing used is the port - pass in the port

casual/casual-test/src/test/groovy/se/laz/casual/test/CasualEmbeddedServerEventTest.groovy
```diff
@@ -0,0 +1,171 @@
...
+            .withPID(pid1)
+            .withExecution(execution1)
+            .withTransactionId(transactionId1)
+            .withPending( pending1 )
+            .withStart( start1 )
+            .withEnd( end1 )
+            .withCode(code1)
+            .withOrder(order1)
+            .build()
+    CountDownLatch countDownLatch = new CountDownLatch( 1 )
```
aliothe, line 54: Nice to use CountDownLatch

casual/casual-test/src/test/groovy/se/laz/casual/test/CasualEmbeddedServerTest.groovy
```diff
@@ -0,0 +1,104 @@
...
+
+    def cleanup()
+    {
+        if ( instance != null && instance.isRunning() )
+        {
+            instance.shutdown()
+        }
+    }
+
+    def "Create then get"()
```
aliothe, line 29: Not really what the test does?

casual/casual-test/src/test/groovy/se/laz/casual/test/CasualEmbeddedServerTest.groovy
```diff
@@ -0,0 +1,104 @@
...
+
+        when:
+        instance.shutdown()
+
+        then:
+        !instance.isRunning()
+        noExceptionThrown()
+    }
+
+    def "Start can only be called once."()
```
aliothe, line 74: It would make sense that start is also idempotent no?

casual-test.md
cklellie: I do somethings if on my machine with many cores, changes `clean build` from 2m to 1m20s, so it's worth it.
Equally, this is meant to be used by others, so it would be good to not enforce constraints upon how they run their tests where possible.

casual/casual-event-api/src/main/java/se/laz/casual/event/CasualServiceCallEventStore.java
cklellie: There is nothing which says it must be static though. Why must the data be stored with the class?

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventStoreFactory.java
cklellie: The design of the event server is free to change at this point. There is no issue with this change, other than it is different. Equally, if you can argue this adds the potential in the future to break out the event server to share between casual instances allowing the domain id to determine where it is retrieved from. 
I prefer this over changing just to a new SPI as i want to reuse the existing code base as must as possible.

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServerConnectionInformation.java
cklellie: Can't remember why i changed this, it was when i was trying to debug why the clients were not getting events initially. Something it popped when i tried to enable it. So i fudged it. I can try and revert it back now i have

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServerConnectionInformation.java
cklellie: yeah i'll try and revert, see previous reply.

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
cklellie: dang it intellij!

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
cklellie: It is idempotent. Idempotency is about the resulting state of the system not the response. 
If you have told it to start multiple times, you have messed up i believe, hence illegal state.

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
cklellie: Yes it can, if you start the embedded server, but don't have the event server enabled, it wont start an event server. In future versions it will be more than just event server that is started when you start the embedded server.

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
cklellie: I have missed eventServerEnabled, thank you. It is a copy builder to make a very similar but different from a concrete instance. I forgot to add the additional components. In future versions there will be likely many more properties to copy over.

casual/casual-test/src/test/groovy/se/laz/casual/test/CasualEmbeddedServerTest.groovy
cklellie: erm, yeah it does, but the `getters` are `is` because they are `boolean`s, the last is a `get`. I can update the test "create then check state" perhaps?

casual/casual-test/src/test/groovy/se/laz/casual/test/CasualEmbeddedServerTest.groovy
cklellie: it is, and i think it should error if you try and start it twice, see previous comment.

casual-test.md
aliothe: What I mean is, when we run clean build today, does that run our unit tests in parallell?

casual/casual-event-api/src/main/java/se/laz/casual/event/CasualServiceCallEventStore.java
aliothe: Besides that the JVM can optimize things it also shows clear intent, for the whole lifespan of our code - there will only be this storage.

casual-test.md
cklellie: No, but it's just `--parallel` on the command :) https://docs.gradle.org/current/userguide/performance.html

casual/casual-event-api/src/main/java/se/laz/casual/event/CasualServiceCallEventStore.java
cklellie: As per discussion we are agreed that in this scenario we are ok with it.

casual/casual-event-server/src/main/java/se/laz/casual/event/server/EventServerConnectionInformation.java
cklellie: Now I remember, it was because i wanted to be able to set this without setting an environment variable.
Currently in the builder it is always read from the env.
So it is only in the builder. I want it to be that if you have not specified anything during the builder setup, then it will run as it currently does with an env read. but if I have specified the setting. I don't want it to be overwritten by the env. Hence i need to use an Object Boolean in the builder only to show whether it has been set or not.

aliothe: Alrighty, so there was a case for it

casual/casual-test/src/main/java/se/laz/casual/test/CasualEmbeddedServer.java
cklellie: After discussion we agreed it was ok to fail fast on multiple starts, as it is more likely user error than desired.

aliothe (approved):

PR #118, opened 2024-06-04
Feature/2.2/reverse inbound share event loop group

aliothe:
* All reverse inbound clients now shares the same event loop group instead of creating a new per connection
  This works the same way as it currently does for outbound connections.
* New root property useEpoll - can also be set via env var CASUAL_USE_EPOLL
  If set to true, epoll is used for outbound, inbound and reverse inbound
* New root property unmanaged - can also be configured via env var CASUAL_UNMANAGED
  If set to true, outbound and reverse inbound runs using a managed executor ( only supported on jboss/wildfly currently)
* Netty version bump to 4.1.110.Final
* The deprecated configuration options re epoll and unmanaged will be removed in a future version

### DIFF
--- a/configuration.md
+++ b/configuration.md
@@ -14,8 +14,8 @@ Casual supports the following configuration options which can be set using envir
 * `CASUAL_CONFIG_FILE` - Set path for casual configuration file. If not provided, default are used.
 * `CASUAL_INBOUND_STARTUP_MODE` - Set mode for inbound startup. Default `immediate`. Alternatives `trigger`, `discover`.
     See [Inbound Startup Configuration](inbound.md#startup-configuration) for more details.
-* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please see `CASUAL_USE_EPOLL` instead.
-* `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please see `CASUAL_USE_EPOLL` instead.
+* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please use `CASUAL_USE_EPOLL` instead.
+* `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please use `CASUAL_USE_EPOLL` instead.
 * `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO, this goes for outbound, inbound and reverse inbound.
 * `CASUAL_INBOUND_STARTUP_INITIAL_DELAY_SECONDS` - Delay the startup of the inbound server. Default is no delay.
 * `CASUAL_UNMANAGED_SCHEDULED_EXECUTOR_SERVICE_POOL_SIZE` - The pool size of casual's scheduled executor service instance. Defaults to 10.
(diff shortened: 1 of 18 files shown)


configuration.md
```diff
@@ -14,10 +14,12 @@ Casual supports the following configuration options which can be set using envir
...
 * `CASUAL_INBOUND_STARTUP_MODE` - Set mode for inbound startup. Default `immediate`. Alternatives `trigger`, `discover`.
     See [Inbound Startup Configuration](inbound.md#startup-configuration) for more details.
-* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
-* `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
+* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please see `CASUAL_OUTBOUND_USE_EPOLL` instead.
+* `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please see `CASUAL_OUTBOUND_USE_EPOLL` instead.
+* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO, this goes for outbound, inbound and reverse inbound.
 * `CASUAL_INBOUND_STARTUP_INITIAL_DELAY_SECONDS` - Delay the startup of the inbound server. Default is no delay.
 * `CASUAL_UNMANAGED_SCHEDULED_EXECUTOR_SERVICE_POOL_SIZE` - The pool size of casual's scheduled executor service instance. Defaults to 10.
+* `CASUAL_UNMANAGED` - If set to true, outbound and reverse inbound will use a managed executor service.
```
TobiasLeo: Is this really correct? Won't it use an `unmanaged` executor if set to `true`?

configuration.md
```diff
@@ -14,10 +14,12 @@ Casual supports the following configuration options which can be set using envir
 * `CASUAL_CONFIG_FILE` - Set path for casual configuration file. If not provided, default are used.
 * `CASUAL_INBOUND_STARTUP_MODE` - Set mode for inbound startup. Default `immediate`. Alternatives `trigger`, `discover`.
     See [Inbound Startup Configuration](inbound.md#startup-configuration) for more details.
-* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
-* `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO.
+* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please see `CASUAL_OUTBOUND_USE_EPOLL` instead.
```
TobiasLeo: This documentation says deprecated, but from what I can see the use of this environment variable has been outright removed. Does anyone currently use this env who will be affected by the change and switch?

configuration.md
aliothe: You are 100% correct, it is the other way around.
If set to true it will not use any managed executor.

configuration.md
aliothe: Hmm, did I remove it by accident?
I'll add it back.

configuration.md
```diff
@@ -14,10 +14,12 @@ Casual supports the following configuration options which can be set using envir
...
+* `CASUAL_INBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO - deprecated, please see `CASUAL_OUTBOUND_USE_EPOLL` instead.
+* `CASUAL_OUTBOUND_USE_EPOLL` - If set to true, Netty uses epoll instead of NIO, this goes for outbound, inbound and reverse inbound.
```
TobiasLeo: This should say something else than `CASUAL_OUTBOUND_USE_EPOLL`? `CASUAL_USE_EPOLL` maybe based on the code changes. Also goes for the deprecation notices which seem to reference the incorrect env

aliothe: It is indeed CASUAL_USE_EPOLL, documentation was not updated correctly - it is now.

TobiasLeo: The commented line (19) should also be updated to `CASUAL_USE_EPOLL` instead of `CASUAL_OUTBOUND_USE_EPOLL`, right?

aliothe: Fixed

TobiasLeo (approved):

PR #122, opened 2024-09-26
Feature/3.2/execution sticky

aliothe:
In the service call API, expose methods, tp(a)call, that additionally accepts a execution id.

casual-caller will make use of this new API when n number of service calls are made in the same transaction and sticky is enabled.
The same execution will then be used for all service calls in the same transaction and it will help with traceability.


TobiasLeo (approved):
Looks good to me

PR #126, opened 2024-10-29
Bugfix/3.2/configuration mode not honoured

cklellie:
Complete refactoring of configuration functionality.
Flattening of the configuration object into a key value store to simplify retrieval.
Simplified handling of order of precedence.
Made all json POJOs package private to prevent them from being used going forward, other than to read the config file into the new ConfigurationStore.

There should be not change in the functionality as a result of this change.
All existing options remain.
The only exception is the fix to the initial bug related to inbound startup mode not honouring the env if provided.

I still need to run a few integration tests, but push this out there so you can review it in advance.


casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfoProvider.java
```diff
@@ -1,25 +1,26 @@
...
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
  * Provides the encoding information for encoding/decoding strings
- * Defaults to UTF-8 but can be set via the environment variable {@code CASUAL_API_FIELDED_ENCODING}
+ * Value determined by the {@link ConfigurationService}.
```
aliothe, line 19: It does not deafult to utf-8 anymore?

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationEnvsReader.java
```diff
@@ -0,0 +1,187 @@
...
+        populateInbound();
+        populateEpoll();
+        populateEventServer();
+        populateUnmanaged();
+        populateOutbound();
+    }
+
+    private void populateOutbound()
+    {
+        storeIntegerIfPresent( ConfigurationOptions.CASUAL_OUTBOUND_MANAGED_EXECUTOR_NUMBER_OF_THREADS );
```
aliothe, line 56: Could have used the storeIfPresent with the correct lambda parameter for all of these instead of indirection to methods that does it.
However, it is a matter of taste.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
```diff
@@ -1,82 +1,149 @@
...
-        this.configuration = init();
+        this.store = init();
     }
 
-    public static final ConfigurationService getInstance()
+    private ConfigurationStore init()
     {
-        return INSTANCE;
-    }
+        ConfigurationStore store = new ConfigurationStore();
```
aliothe: shadowing, rename

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
```diff
@@ -1,82 +1,149 @@
...
+        ConfigurationEnvsReader envsReader = new ConfigurationEnvsReader( store );
+
+        envsReader.populateConfigFileEnv();
+
+        String configurationFile = store.get( ConfigurationOptions.CASUAL_CONFIG_FILE );
+        if( configurationFile != null && !configurationFile.isBlank() )
+        {
+            ConfigurationFileReader fileReader = new ConfigurationFileReader( store );
+            fileReader.populateStoreFromFile( configurationFile );
+        }
```
aliothe, line 46: Could this not be handled by the ConfigurationFileReader?

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
```diff
@@ -1,82 +1,149 @@
...
-    private Configuration buildConfigurationFromFile( String file )
+    private void fixEpoll( ConfigurationStore store )
     {
-        try
+        boolean rootEpoll = store.get( ConfigurationOptions.CASUAL_USE_EPOLL );
+        if( rootEpoll )
         {
-            return JsonProviderFactory.getJsonProvider().fromJson( new FileReader( file ), Configuration.class );
+            store.put( ConfigurationOptions.CASUAL_OUTBOUND_USE_EPOLL, true );
+            store.put( ConfigurationOptions.CASUAL_INBOUND_USE_EPOLL, true );
```
aliothe, line 81: In the end we want to get rid of these deprecated env vars, not now but sometime in the future.
Will that be easy to do with this solution?

aliothe: Maybe we should log if somebody does use CASUAL_OUTBOUND_USE_EPOLL and/or CASUAL_INBOUND_USE_EPOLL?
So people will notice that they are deprecated and will be removed in the future?

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
aliothe, line 1: I would prefer to see the public methods before the private methods

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationStore.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Store for all configuration options.
+ */
+public class ConfigurationStore
+{
+    private final Map<ConfigurationOption<?>, Object> data;
+
+    public ConfigurationStore()
+    {
+        data = new HashMap<>();
```
aliothe, line 21: ConcurrentHashMap not needed?

casual/casual-api/src/main/java/se/laz/casual/config/Mode.java
```diff
@@ -1,23 +1,21 @@
...
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * Inbound Startup Mode.
+ */
 public enum Mode
 {
-    @SerializedName(Constants.IMMEDIATE)
     IMMEDIATE( Constants.IMMEDIATE ),
```
aliothe, line 17: Nice to get rid of these! :)

casual/casual-api/src/main/java/se/laz/casual/config/ReverseInbound.java
```diff
@@ -1,87 +1,129 @@
...
-        this.size = size;
-        this.maxConnectionBackoffMillis = maxConnectionBackoffMillis;
-    }
+    private final String host;
+    private final Integer port;
+    private final Integer size;
+    private final Long maxConnectionBackoffMillis;
 
-    public static ReverseInbound of(Address address, int size, long maxBackoffMillis)
+    public ReverseInbound( Builder builder )
```
aliothe, line 20: Only used from the builder, make it private

casual/casual-api/src/main/java/se/laz/casual/config/json/Configuration.java
```diff
@@ -0,0 +1,177 @@
...
+{
+    private final Inbound inbound;
+    private final Domain domain;
+    private final Outbound outbound;
+    private final List<ReverseInbound> reverseInbound;
+    private final EventServer eventServer;
+    private final Boolean useEpoll;
+    private final Boolean unmanaged;
+
+    public Configuration( Builder builder )
```
aliothe: Only used from the builder, make it private

casual/casual-api/src/main/java/se/laz/casual/config/json/Configuration.java
```diff
@@ -0,0 +1,177 @@
...
+
+        public Builder withOutbound( Outbound outbound )
+        {
+            this.outbound = outbound;
+            return this;
+        }
+
+        public Builder withReverseInbound( List<ReverseInbound> reverseInbounds )
+        {
+            this.reverseInbound = reverseInbounds;
```
aliothe: Maybe be defensive and copy the values to a new list that we own here?

casual/casual-api/src/main/java/se/laz/casual/config/json/Configuration.java
```diff
@@ -0,0 +1,177 @@
...
+        return new Builder().withInbound( src.getInbound() )
+                .withOutbound( src.getOutbound() )
+                .withReverseInbound( src.getReverseInbound() )
+                .withEventServer( src.getEventServer() )
+                .withUseEpoll( src.getUseEpoll() )
+                .withUnmanaged( src.getUnmanaged() )
+                .withDomain( src.getDomain() );
+    }
+
+    public static final class Builder
```
aliothe, line 121: Is it ok that the parameters are null?

casual/casual-api/src/main/java/se/laz/casual/config/json/Address.java
```diff
@@ -0,0 +1,94 @@
...
+    {
+        return new Builder();
+    }
+
+    public static Builder newBuilder( Address src )
+    {
+        return new Builder().withHost( src.host ).withPort( src.port );
+    }
+
+    public static final class Builder
```
aliothe, line 72: Is it ok that the parameters are null?

casual/casual-api/src/main/java/se/laz/casual/config/json/Domain.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * Holds the fixed values for id and name for the whole domain.
+ */
+final class Domain
+{
+    private final UUID id;
```
aliothe, line 17: Should be of type DomainId

casual/casual-api/src/main/java/se/laz/casual/config/json/EventServer.java
```diff
@@ -3,36 +3,36 @@
...
+package se.laz.casual.config.json;
 
 import java.util.Objects;
 
-public final class EventServer
+final class EventServer
 {
-    private final int portNumber;
-    private final boolean useEpoll;
+    private final Integer portNumber;
```
aliothe, line 12: Can the port# be null?

casual/casual-api/src/main/java/se/laz/casual/config/json/Inbound.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.util.Objects;
+
+class Inbound
+{
+    private final Startup startup;
+    private final Boolean useEpoll;
+    private final Long initialDelay;
+
+    public Inbound( Builder builder )
```
aliothe: only used in builder, make it private

casual/casual-api/src/main/java/se/laz/casual/config/json/Mode.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package se.laz.casual.config.json;
+
+import com.google.gson.annotations.SerializedName;
+import se.laz.casual.config.Mode.Constants;
+
+enum Mode
+{
+    @SerializedName( Constants.IMMEDIATE)
```
aliothe, line 14: Can we get rid of SerializedName?

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationOptionTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+        new ConfigurationOption<Object>( null )
+
+        then:
+        thrown NullPointerException
+    }
+
+    def "equals and hashcode"()
+    {
+        when:
+        ConfigurationOption<String> instance2 = new ConfigurationOption<>( "mykey" )
```
aliothe: Use name

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
```diff
@@ -6,287 +6,209 @@
...
+              .execute( {
+                  ConfigurationService.reload(  )
+              } )
+      boolean epoll = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_USE_EPOLL )
+
+      then:
+      epoll
+   }
+
+    def "Get configuration where file not found, throws CasualRuntimeException."()
```
aliothe: Throws ConfigurationException

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
```diff
@@ -6,287 +6,209 @@
...
-        Configuration actual = instance.getConfiguration()
+        String actual = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_NAME )
 
         then:
         actual == expected
     }
 
     @Unroll
-    def "Get configuration where file is provided, returns matching configuration."()
+    def "Get configuration where only file env is provided, no other envs, returns matching configuration."()
```
aliothe, line 42: Add at least #file so that we see in the test case what it is running

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
```diff
@@ -6,287 +6,209 @@
...
+                .execute( {
+                    ConfigurationService.reload(  )
+                } )
 
+        then:
+        thrown ConfigurationException
+    }
+
+    @Unroll
+    def "Get configuration where file empty and env mode is provided, returns matching configuration."()
```
aliothe, line 91: At least #env to get the difference per unit test name

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
```diff
@@ -6,287 +6,209 @@
...
-        'casual-config-outbound-epoll.json'                   || 'java:comp/DefaultManagedExecutorService'                           || 0               || false     || true
+        env                      || mode                         | services
+        Mode.Constants.IMMEDIATE || Mode.IMMEDIATE               | []
+        Mode.Constants.TRIGGER   || Mode.TRIGGER                 | [Mode.Constants.TRIGGER_SERVICE]
+        Mode.Constants.DISCOVER  || Mode.DISCOVER                | []
+        ''                       || INBOUND_STARTUP_MODE_DEFAULT | []
     }
 
-    def 'default outbound config, no file'()
+    def "Unmanaged confusion for default and settings."()
```
aliothe, line 115: unroll and #file

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
```diff
@@ -0,0 +1,171 @@
...
+        when:
+        instance.populateStoreFromFile( file )
+
+        then:
+        store.get( ConfigurationOptions.CASUAL_DOMAIN_NAME ) == ""
+        store.get( ConfigurationOptions.CASUAL_INBOUND_STARTUP_MODE ) == se.laz.casual.config.Mode.DISCOVER
+        store.get( ConfigurationOptions.CASUAL_INBOUND_STARTUP_SERVICES ) == ["service1","service2"]
+    }
+
+    def "Read file domain info, store updated"()
```
aliothe, line 56: unroll and #filename so we can see the difference in the test method names

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
```diff
@@ -0,0 +1,171 @@
...
+
+        where:
+        filename                                      || expected
+        "casual-config-domain-info.json"              || "casual-java-test"
+        "casual-config-domain-info-missing.json"      || DOMAIN_NAME_DEFAULT
+        "casual-config-domain-info-name-missing.json" || DOMAIN_NAME_DEFAULT
+        "casual-config-domain-info-null.json"         || DOMAIN_NAME_DEFAULT
+    }
+    
+    def "Read file event server, store updated"()
```
aliothe, line 75: unroll and #filename so we can see the difference in the test method names

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
```diff
@@ -0,0 +1,171 @@
...
+
+        where:
+        filename                                    || enabled                      | port                      | epoll             | quietPeriod                   | timeout
+        "casual-config-empty.json"                  || EVENT_SERVER_ENABLED_DEFAULT | EVENT_SERVER_PORT_DEFAULT | USE_EPOLL_DEFAULT | SHUTDOWN_QUIET_PERIOD_DEFAULT | SHUTDOWN_TIMEOUT_DEFAULT
+        "casual-config-event-server-use-epoll.json" || true                         | 6699                      | true              | SHUTDOWN_QUIET_PERIOD_DEFAULT | SHUTDOWN_TIMEOUT_DEFAULT
+        "casual-config-event-server-no-epoll.json"  || true                         | 9966                      | USE_EPOLL_DEFAULT | SHUTDOWN_QUIET_PERIOD_DEFAULT | SHUTDOWN_TIMEOUT_DEFAULT
+        "casual-config-event-server-shutdown.json"  || true                         | 5987                      | true              | 100                           | 200
+    }
+
+    def "Read file inbound, store updated"()
```
aliothe, line 98: unroll and #filename so we can see the difference in the test method names

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
```diff
@@ -0,0 +1,171 @@
...
+        where:
+        filename                                   || mode                                | epoll             | delay                                 | services
+        "casual-config-inbound-immediate.json"     || se.laz.casual.config.Mode.IMMEDIATE | USE_EPOLL_DEFAULT | INBOUND_STARTUP_INITIAL_DELAY_DEFAULT | INBOUND_START_SERVICES_DEFAULT
+        "casual-config-inbound-trigger.json"       || se.laz.casual.config.Mode.TRIGGER   | USE_EPOLL_DEFAULT | INBOUND_STARTUP_INITIAL_DELAY_DEFAULT | INBOUND_START_SERVICES_DEFAULT
+        "casual-config-inbound-discover.json"      || se.laz.casual.config.Mode.DISCOVER  | USE_EPOLL_DEFAULT | INBOUND_STARTUP_INITIAL_DELAY_DEFAULT | ["service1", "service2"]
+        "casual-config-inbound-epoll.json"         || INBOUND_STARTUP_MODE_DEFAULT        | true              | INBOUND_STARTUP_INITIAL_DELAY_DEFAULT | INBOUND_START_SERVICES_DEFAULT
+        "casual-config-inbound-initial-delay.json" || INBOUND_STARTUP_MODE_DEFAULT        | USE_EPOLL_DEFAULT | 30L                                   | INBOUND_START_SERVICES_DEFAULT
+    }
+
+    def "Read file outbound, store updated"()
```
aliothe, line 121: unroll and #filename so we can see the difference in the test method names

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
```diff
@@ -0,0 +1,171 @@
...
+        "casual-config-outbound-epoll.json"                   || OUTBOUND_MANAGED_EXECUTOR_SERVICE_NAME_DEFAULT          | true              | EXECUTOR_NUMBER_OF_THREADS_DEFAULT | UNMANAGED_DEFAULT
+        "casual-config-outbound-executorName-missing.json"    || OUTBOUND_MANAGED_EXECUTOR_SERVICE_NAME_DEFAULT          | USE_EPOLL_DEFAULT | 10                                 | UNMANAGED_DEFAULT
+        "casual-config-outbound-null.json"                    || OUTBOUND_MANAGED_EXECUTOR_SERVICE_NAME_DEFAULT          | USE_EPOLL_DEFAULT | EXECUTOR_NUMBER_OF_THREADS_DEFAULT | UNMANAGED_DEFAULT
+        "casual-config-outbound-numberOfThreads-missing.json" || "java:comp/env/concurrent/casualManagedExecutorService" | USE_EPOLL_DEFAULT | EXECUTOR_NUMBER_OF_THREADS_DEFAULT | UNMANAGED_DEFAULT
+        "casual-config-outbound-unmanaged.json"               || OUTBOUND_MANAGED_EXECUTOR_SERVICE_NAME_DEFAULT          | USE_EPOLL_DEFAULT | EXECUTOR_NUMBER_OF_THREADS_DEFAULT | true
+    }
+
+    def "Read file reverse inbound, store updated"()
```
aliothe, line 145: unroll and #filename so we can see the difference in the test method names

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
```diff
@@ -71,10 +72,11 @@ public void domainConnectRequest(CasualNWMessage<CasualDomainConnectRequestMessa
     {
         log.finest(() -> "domainConnectRequest(). " + PrettyPrinter.format(message.getCorrelationId(), message.getMessage().getExecution()) + message );
 
-        Domain domain = ConfigurationService.getInstance().getConfiguration().getDomain();
+        String domainName = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_NAME );
+        UUID domainId = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_ID ).getId();
```
aliothe, line 76: Should be DomainId

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
```diff
@@ -88,8 +90,9 @@ public void domainDiscoveryRequest(CasualNWMessage<CasualDomainDiscoveryRequestM
     {
         log.finest(() -> "domainDiscoveryRequest(). " + PrettyPrinter.format(message.getCorrelationId(), message.getMessage().getExecution()) + message);
 
-        Domain domain = ConfigurationService.getInstance().getConfiguration().getDomain();
-        CasualDomainDiscoveryReplyMessage reply = CasualDomainDiscoveryReplyMessage.of( message.getMessage().getExecution(), domain.getId(), domain.getName() );
+        String domainName = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_NAME );
+        UUID domainId = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_ID ).getId();
```
aliothe, line 94: DomainId

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/ServiceCallWorkListener.java
```diff
@@ -102,7 +103,7 @@ ServiceCallEventPublisher getEventPublisher()
     {
         if(eventPublisher == null)
         {
-            UUID domainId = ConfigurationService.getInstance().getConfiguration().getDomain().getId();
+            UUID domainId = ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_ID ).getId();
```
aliothe, line 106: DomainId

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
```diff
@@ -76,29 +77,33 @@ public class CasualResourceAdapter implements ResourceAdapter, ReverseInboundLis
...
     {
         //JCA requires ResourceAdapter has a no arg constructor.
         //It is also not possible to inject with CDI on wildfly only ConfigProperty annotations.
-        configurationService = ConfigurationService.getInstance();
-        log.info(() -> "casual jca configuration: " + configurationService.getConfiguration());
-        configurationService.getConfiguration().getEventServer().ifPresent(config -> {
-            log.info("starting event server with config: " + config);
+
+        log.info(() -> "casual jca configuration:" );
+        log.info( ConfigurationService.log() );
```
aliothe, line 85: lambda, be lazy - might not be info level so no need to evaluate if it is not

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
```diff
@@ -76,29 +77,33 @@ public class CasualResourceAdapter implements ResourceAdapter, ReverseInboundLis
...
-                    .withShutdownTimeout( config.getShutdown().getTimeout() )
-                    .withShutdownQuietPeriod( config.getShutdown().getQuietPeriod() )
-                    .build(), configurationService.getConfiguration().getDomain().getId() );
-            log.info("event server started at port: " + config.getPortNumber());
+                    .withUseEpoll( ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_EVENT_SERVER_USE_EPOLL ) )
+                    .withPort( ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_EVENT_SERVER_PORT ) )
+                    .withShutdownTimeout( ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_EVENT_SERVER_SHUTDOWN_TIMEOUT_MILLIS ) )
+                    .withShutdownQuietPeriod( ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_EVENT_SERVER_SHUTDOWN_QUIET_PERIOD_MILLIS ) )
+                    .build(), ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_DOMAIN_ID ).getId() );
+            log.info("event server started at port: " + ConfigurationService.getConfiguration( ConfigurationOptions.CASUAL_EVENT_SERVER_PORT ) );
```
aliothe: lambda

casual/casual-jca/src/main/java/se/laz/casual/jca/discovery/CasualDiscoveryCaller.java
```diff
@@ -47,11 +47,11 @@ public static CasualDiscoveryCaller of(CasualManagedConnection managedConnection
     public DiscoveryReturn discover(UUID corrid, List<String> serviceNames, List<String> queueNames)
     {
         LOG.finest(() -> "issuing domain discovery, corrid: " + PrettyPrinter.casualStringify(corrid) + " service names: " + serviceNames + " queue names: " + queueNames);
-        Domain domain = ConfigurationService.getInstance().getConfiguration().getDomain();
```
aliothe, line 50: The Domain type is useful, a casual java domain - doman name(String) and domain id(DomainId)
I do not see the need to remove it still

casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfoProvider.java
cklellie: Yes, but as stated it is now determined by the ConfigurationService. So removed from javadoc here.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationStore.java
cklellie: There is no concurrent writing occuring. Only updated during initial startup which is single threaded. Only reads after that. So no, i don't see the need for ConcurrentHashMap here.

casual/casual-api/src/main/java/se/laz/casual/config/ReverseInbound.java
cklellie: So weird, I thought i had done all these! Thanks.

casual/casual-api/src/main/java/se/laz/casual/config/json/Domain.java
cklellie: No, as that would break the json marshalling, it would become `{ "domain": { "id": { "id": "" } } }`

casual/casual-api/src/main/java/se/laz/casual/config/json/EventServer.java
cklellie: This is a json pojo only, if it is not present I do not want any kind of defaulting within the json pojo either in the builder as this is not invoked by gson or through primatives as i want final fields and as i want all defaults controlled by the ConfigurationDefaults class only.
This was part of the issue when refactoring, there were lots of different locations where the default could be determined. I have now consolidated that into one place. As a result we use Objects rather than primatives, but only in the package private json POJOs so that we can handle these scenarios and get the correct default values set when they are in fact missing in the json i.e. null.

casual/casual-api/src/main/java/se/laz/casual/config/json/Configuration.java
cklellie: yes, it's json.

casual/casual-api/src/main/java/se/laz/casual/config/json/Address.java
cklellie: yes, it's json.

casual/casual-api/src/main/java/se/laz/casual/config/json/Mode.java
cklellie: no, it is json so needed still.

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
cklellie: We are using spock version 2, not needed anymore.

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
cklellie: We are using spock version 2, not needed anymore.

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
cklellie: We are using spock version 2, not needed anymore.

casual/casual-api/src/test/groovy/se/laz/casual/config/json/ConfigurationFileReaderTest.groovy
cklellie: We are using spock version 2, not needed anymore.

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
cklellie: But I want the value as a UUID to use 2 lines later. So I am preparing it here for that.

casual/casual-api/src/test/groovy/se/laz/casual/config/ConfigurationServiceTest.groovy
aliothe: Cool, so it automagically uses the names and values from the datatable, did not know that! :)

casual/casual-jca/src/main/java/se/laz/casual/jca/discovery/CasualDiscoveryCaller.java
cklellie: If it was useful it would have been used as a parameter into a method somewhere, but it is not, anywhere, at all. Equally DomainId, though i have used that, it does not really provide any value add. If you go too far down this route you would have a class for every field, HostName, PortNumber, ServiceName, QueueName. You have to be able to use primatives at some point.
The only places Domain was used was to retrieve it from the json, which in turn didn't use a DomainId, because it can't becase that is not how the json structure is provided as i mentioned earlier. The usages of Domain object always immediately just split the object up into it's string and UUID components to send into other classes which required String and UUID.

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
cklellie: Return type of `ConfigurationService.log()` is a `Supplier<String>`, so it is already lazy.

casual/casual-jca/src/main/java/se/laz/casual/jca/discovery/CasualDiscoveryCaller.java
aliothe: Sure, we could go with primitives everywhere - that would be horrible though.
Speaking of DomainId, it is also used in casual-caller, as DomainId - it does not have to know what kind of stuff is inside of it.
As for Domain, it is a mental construct as well, a java Domain from our point of view consists of a name ( should probably be a default and not empty if not set) and a unique DomainId.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
cklellie: yeah, but they are all related to the constructor here so i think  this more readable. Similarly, if i have public methods that call private methods i would group them below each public, not a block of public and then a seperate block of all privates. As if i want to read the functions in one go it's all there in front of me with less scrolling.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
cklellie: It is. What do you mean?

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationEnvsReader.java
cklellie: But then i would have to repeat the lambda/method reference on each usage, this is a convinience method to reduce duplication for each invocation for Integers, similarly for Booleans, Longs etc etc. Which also then makes it more readable as you can see from the  method name, which you read first, what it is doing.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
cklellie: Yeah, not sure, I'll make an issue for this and we can discuss options for a later release.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
aliothe: So we have a mix of private/public method sections all over this code base?

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationEnvsReader.java
aliothe: Yeah, as I said - it is a matter of taste.

casual/casual-api/src/main/java/se/laz/casual/config/ConfigurationService.java
cklellie: Probably, another example you made, which follows similar rationale I would argue:
casual/casual-network/src/main/java/se/laz/casual/network/inbound/CasualServer.java
Then all the Constructors which use builders and private, I don't want them to be pushed down.

casual/casual-jca/src/main/java/se/laz/casual/jca/discovery/CasualDiscoveryCaller.java
cklellie: yes, but as I am saying, it can go too far the other way too.
Equally the "domain" concept for casual is already well used and means more than just those two things.
Anyway, I think we have entered bike shed territory here, so lets just park this.

aliothe (approved):

PR #127, opened 2024-11-01
trid unique per resource, not global

aliothe:
For wildfly, the trid should be unique per resource - not global as it is for wls as wls handles it differently.

### DIFF
--- a/casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceManager.java
+++ b/casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceManager.java
@@ -8,6 +8,8 @@ package se.laz.casual.jca;
 
 
 import javax.transaction.xa.Xid;
+import java.util.Collections;
+import java.util.Set;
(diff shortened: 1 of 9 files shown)


casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceManager.java
```diff
@@ -17,38 +19,47 @@
...
         {
-            throw new CasualResourceAdapterException("xid: " + xid + " already stored");
+            throw new CasualResourceAdapterException("xid: " + xid + " already stored for domain: " + domainId);
         }
-        pendingRequests.put(xid, true);
+        pendingRequests.computeIfAbsent(domainId, k -> ConcurrentHashMap.newKeySet()).add(xid);
     }
 
-    public void remove(final Xid xid)
+    public synchronized void remove(DomainId domainId, final Xid xid)
```
cklellie, line 47: Should you also synchronize put too then?

cklellie (approved):

PR #130, opened 2024-11-25
Bugfix/3.2/inbound discovery with custom service handlers

cklellie:
Utilise all available service handlers to determine when services are available on startup with mode DISCOVER.
This is to ensure that custom service handlers are not ignored, as it currently the case.

### DIFF
--- a/casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/service/ServiceHandler.java
+++ b/casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/service/ServiceHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 - 2018, The casual project. All rights reserved.
+ * Copyright (c) 2017 - 2024, The casual project. All rights reserved.
  *
  * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
  */
@@ -18,7 +18,7 @@ public interface ServiceHandler extends Prioritisable
      * service of this name.
      *
      * Differs from {@link #isServiceAvailable} in that a
-     * service could be registered  but unavailable,
+     * service could be registered and possible to handle but unavailable,
      * in which case this method would return true,
      * but {@link #isServiceAvailable} would return false.
      *
@@ -28,8 +28,15 @@ public interface ServiceHandler extends Prioritisable
     boolean canHandleService( String serviceName );
 
     /**
-     * Determine if the provided service is available
-     * in this handler.
+     * Determine if the provided service is available to receive requests
+     * through this handler.
+     *
+     * Differs from {@link #canHandleService} in that a
+     * service may not yet be available/ready to receive requests, but
+     * this handler can handle the service when it is available.
+     * In which case this method would return false,
+     * but {@link #canHandleService} would return true.
+     *
      * @param serviceName name
      * @return if the service is available.
      */
--- a/casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
+++ b/casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
@@ -10,11 +10,12 @@ import jakarta.resource.spi.work.Work;
 import se.laz.casual.config.ConfigurationOptions;
 import se.laz.casual.config.ConfigurationService;
 import se.laz.casual.jca.InboundStartupException;
-import se.laz.casual.jca.inbound.handler.service.casual.CasualServiceRegistry;
+import se.laz.casual.jca.inbound.handler.service.ServiceHandler;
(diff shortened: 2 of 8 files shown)


casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/service/ServiceHandler.java
```diff
@@ -18,7 +18,7 @@ public interface ServiceHandler extends Prioritisable
      * service of this name.
      *
      * Differs from {@link #isServiceAvailable} in that a
-     * service could be registered  but unavailable,
+     * service could be registered and possible to handle but unavailable,
```
aliothe, line 21: What does this mean, possible to handle but unavailable?
What can be handled if it is unavailable?

casual/casual-jca/src/main/java/se/laz/casual/jca/work/StartInboundServerWork.java
```diff
@@ -113,14 +114,25 @@ private void logWaitingForStartupServices(Set<String> remaining)
...
-        Map<Boolean,Set<String>> found = remaining.stream()
-                .collect( Collectors.partitioningBy( registry::hasServiceEntry, Collectors.toSet() ) );
-        for( String foundService: found.get( true ) )
+        Set<String> notFound = new HashSet<>( remaining );
+        for( String find: remaining )
         {
-            log.info( ()-> "Startup service registered: " + foundService );
+            try
+            {
+                ServiceHandler handler = ServiceHandlerFactory.getHandler( find );
```
aliothe, line 122: Nice to use the code we already have

casual/casual-jca/src/test/groovy/se/laz/casual/jca/work/StartInboundServerWorkTest.groovy
```diff
@@ -131,20 +130,19 @@ class StartInboundServerWorkTest extends Specification
...
 
         when:
         completionService.submit( { instance.run(  ) } )
         completionService.submit( {
-            for( CasualServiceEntry entry: entries )
+            for( String service: serviceNames )
             {
-                registry.register( entry )
+                testHandler.makeServiceAvailable( service )
                 Thread.sleep( 1010 )
```
aliothe, line 146: Didn't know we had sleeps in our test code, naughty

inbound.md
```diff
@@ -34,16 +34,25 @@ public class EchoServiceImpl implements EchoService
...
-``` 
-During application deployment Casual JCA application will `@Observe` the deployment of the `@CasualService` and 
+```
+
+All available `ServiceHandler` implementation are used to respond to domain discovery requests by calling the`isServiceAvailable` method. 
+It is therefore the responsibility of the `ServiceHandler` implementations to ensure it responds correctly to these domain discovery requests and
+dispatch incoming requests to the appropriate deployed application.
+
+As new applications containing casual inbound services are deployed, the domain discovery requests must respond accordingly. Therefore all
+`ServiceHandler` implementations must monitor the deployment of applications containing Casual Services over time.
```
aliothe: Hmm, 3rd party service handlers surely does not have to be bothered with CasualService's?

versions.gradle
```diff
@@ -7,7 +7,7 @@
 
 // casual-java group and version
 group = 'se.laz.casual'
-version = '3.2.47'
+version = '3.2.48'
 
 def netty_version = '4.1.114.Final'
```
aliothe, line 12: Maybe bump to https://netty.io/news/2024/11/12/4-1-115-Final.html as well?

aliothe (approved):

casual/casual-inbound-handler-api/src/main/java/se/laz/casual/jca/inbound/handler/service/ServiceHandler.java
cklellie: That it can handle this service if it was up. It is a way to determine dispatch to the service handler independently from the services availability. Idea being that canHandle should be a simpler faster check.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/work/StartInboundServerWorkTest.groovy
cklellie: yeah, i see we have a fixed sleep in the check for available services of 1000. then with this sleep in test being 10ms more, jikes!

inbound.md
cklellie: I think this is what i was talking about before, all services you have for inbound are "casual services", but only a subset are annotated with `@CasualService`.
Perhaps the clarification here is "containing Services available to inbound calls via casual.", or "their services". I don't know feels a bit unclear still.

versions.gradle
cklellie: Is it not better to have bumps to deps as being seperate prs? Keeping each PR focused?

PR #138, opened 2024-12-06
Feature/3.2/bundle gson

aliothe:
Make sure that gson is not needed in the global module since this polluted all deployed applications with that specific gson implementation.

Two new modules:
* casual-json-provider - provides our JsonProvider implementation as well as the dependency on our gson version
* casual-api-spi-impl - Module that wraps all our spi modules, currently only casual-json-provider. This is the module that other applications should depend on.


casual/casual-event-client/build.gradle
```diff
@@ -19,7 +19,7 @@ dependencies {
     
     testImplementation project(':casual:casual-api')
     testImplementation project(':casual:casual-event-api')
-    testImplementation libs.gson
+    testRuntimeOnly project(':casual:casual-api-spi-impl')
```
cklellie, line 23: This will mean that clients consuming the event client will not receive this package and have to add this in their gradle.
Do we not want to have this as implementation instead too?
With api, it is not 100% you will need to have a json thing, but with event client it doesn't work without it.

casual/casual-json-provider/build.gradle
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2024, The casual project. All rights reserved.
+ *
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+
+//-*- coding: utf-8 -*-
+plugins {
```
cklellie: module naming - `casual-json-provider-gson` ?

cklellie (approved):

PR #141, opened 2024-12-10
Feature/3.2/code needed for conversation support in casual caller

aliothe:
Instead of returning a Conversation we now return a TpConnectReturn which potentially wraps a conversation and a connection as well.
It is auto closable and should be used in a try with resources context.

This code will be used in casual caller to support conversation as well.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/api/CasualConversationApi.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/CasualConversationApi.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 - 2018, The casual project. All rights reserved.
+ * Copyright (c) 2017 - 2024, The casual project. All rights reserved.
  *
  * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
  */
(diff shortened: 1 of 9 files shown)


versions.gradle
```diff
@@ -7,7 +7,7 @@
 
 // casual-java group and version
 group = 'se.laz.casual'
-version = '3.2.50'
+version = '3.2.51'
```
cklellie: Make this a version `3.3.0` to reflect the change we have made to the behaviour.

casual/casual-api/src/main/java/se/laz/casual/api/CasualConversationApi.java
```diff
@@ -0,0 +1,48 @@
...
+     * The conversation is only available if {@link TpConnectReturn#getErrorState()} equals to {@link se.laz.casual.api.flags.ErrorState#OK}
+     * - if not, you should fail the call with the provided error.
+     *
+     * Note:
+     * Always issue the call using try-with-resources to make sure that nothing leaks
+     *
+     * @param serviceName - the service name
+     * @param data - initial data to be sent
+     * @param flags - the flags
+     * @return A TpConnectReturn with a conversation if {@link TpConnectReturn#getErrorState()} equals to {@link se.laz.casual.api.flags.ErrorState#OK}
```
cklellie, line 38: Add information about which things can be `@throws`

cklellie (approved):

PR #142, opened 2024-12-10
3.2 back into dev, dev should be default going forward

aliothe:
Merging 3.2 back into dev so that we can make dev default branch going forward.


cklellie (approved):
Just make sure you merge it correctly ;-)

PR #157, opened 2025-03-19
chore(changelog): initial auto tagging and changelog updating

aliothe:
After this, all commits are required to follow:
https://www.conventionalcommits.org/en/v1.0.0/

The squash message that is created is the same as for the PR, you are *NOT* allowed to change this.
Thus create a very good conventional commit message for the PR.
If you change it before squash merging, it will merge it but there is an extra check that will fail and email that it failed - thus it would then have to be reverted.


CHANGELOG.md
```diff
@@ -0,0 +1,1298 @@
+# Changelog
+This is the changelog for *casual java* and all changes are listed in this document.
+
+
+## [3.3.1] - 2025-01-13
+- Feature/3.2/unmarshall all null values (#143)
```
TobiasLeo: Would be really nice if issue refs were hyperlinked like in the main casual repo. It seems to generate lines like `- xatmi: build-server uses [0,1] as values for service visibility ([#506](https://github.com/casualcore/casual/issues/506))`

aliothe: That should be easy enough to add, would be by convention then.
If you have a #[num*] then we'll treat that as an issue and can insert the link.

cklellie: I believe - need to test this myself to check - but this should already work with just the # and a number. See the second link below.
So if you view the markdown file on github ui it will work is my understanding from the documentation.

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#referencing-issues-and-pull-requests

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls

.github/workflows/release-and-update-changelog.yml
```diff
@@ -0,0 +1,52 @@
+name: Release and Update Changelog
+
+on:
+  push:
+    branches:
+      - feature/not-dev
```
cklellie: dev?

.github/workflows/release-and-update-changelog.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  release-and-update:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
```
cklellie: why are we implementing this check?

.github/workflows/release-and-update-changelog.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Update the changelog
+        run: |
+          git config user.name "GitHub Action"
+          git config user.email "action@github.com"
+          python scripts/update_changelog.py
+
+      - name: Commit and push changelog changes
+        run: |
+          if ! git diff --quiet CHANGELOG.md; then
```
cklellie, line 36: How will the changelog already be up to date? Is that if the push did not conform?

scripts/create-initial-changelog.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "- $commit_message" >> $filename
+
+    # Remove leading '*' and whitespace
+    modified_full_commit_message=$(echo "$full_commit_message" | sed 's/^[*[:space:]]*//')
+    
+    [ -n "$modified_full_commit_message" ] && echo "- $modified_full_commit_message" >> $filename
+    echo "" >> $filename
+done
+
+dos2unix $filename
```
cklellie: why is this needed, will it not be run on the same action runner?

scripts/create-initial-changelog.sh
```diff
@@ -0,0 +1,38 @@
...
+    
+    # Get the commit date in YYYY-MM-DD format
+    commit_date=$(git log -1 --pretty=format:"%cd" --date=short "$commit_hash")
+    
+    # Write the tag and commit details to the changelog
+    echo "## [$tag] - $commit_date" >> $filename
+    echo "- $commit_message" >> $filename
+
+    # Remove leading '*' and whitespace
+    modified_full_commit_message=$(echo "$full_commit_message" | sed 's/^[*[:space:]]*//')
```
cklellie: Why remove leading '*'?
For example in this commit:
https://github.com/casualcore/casual-java/commit/62ce53a7c299c113dd5efd1bce168a7ba229893c
I have a list of changes - this would make the first items not part of the list.

scripts/update_changelog.py
```diff
@@ -0,0 +1,46 @@
+#-*- coding: utf-8-unix -*-
+import re
+import subprocess
+from datetime import datetime
+
+# Function to extract version from versions.gradle
+def get_version_from_gradle():
```
cklellie: Why are you duplicating code?

scripts/update_changelog.py
```diff
@@ -0,0 +1,46 @@
...
+try:
+    with open("CHANGELOG.md", "r") as f:
+        changelog = f.read()
+except FileNotFoundError:
+    changelog = "# Changelog\n"
+
+# Find or create the version section
+version_section = f"## [{version}] - {commit_date}"
+if version_section not in changelog:
+    changelog = changelog.replace("This is the changelog for *casual java* and all changes are listed in this document.\n", f"This is the changelog for *casual java* and all changes are listed in this document.\n\n{version_section}\n", 1)
```
cklellie: Use a variable for the string? Potentially have it shared between all three locations including the create-initial-changelog.sh - wait - why is that a bash, but this is now python. Do we need python / both?!

scripts/validate_pr_title.py
```diff
@@ -0,0 +1,20 @@
...
+import sys
+
+# Get the PR title from an environment variable
+pr_title = os.getenv("PR_TITLE", "").strip()
+
+# Define the conventional commit pattern
+pattern = r"^(feat|fix|docs|style|refactor|test|chore)(\(.+\))?: .+"
+
+# Validate the PR title
+if not re.match(pattern, pr_title):
```
cklellie: This function is near identical to the validate_commit_message.py - just a different input and error message. 
Why duplication again?

.github/workflows/release-and-update-changelog.yml
```diff
@@ -0,0 +1,52 @@
...
+            git config user.name "GitHub Action"
```
cklellie, line 37: Why not reuse the same action step - rather than having to repeat the git config?

.github/workflows/validate-pr-title.yml
```diff
@@ -0,0 +1,24 @@
+name: Validate PR Title
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
+    branches:
+      - feature/not-dev
```
cklellie: dev?

.github/workflows/validate-pr-title.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+
+      - name: Validate PR title
```
cklellie, line 21: I think it would be very useful to see what the entry will be in the changelog when you create/update the pr.
Just to sanity check what will be added.
Could we have the validation also output what the log entry will be to stdout, so if you are interested you can check the output of the action?

.gitignore
```diff
@@ -3,6 +3,7 @@
 javadoc/
 docs/source/casual-docs/
 docs/source/version
+.qodo/
```
cklellie, line 6: ? codium? Riiiiight, now it clicks what's up with this PR!

CHANGELOG.md
aliothe: https://github.com/casualcore/casual-java/blob/feature/changelog/CHANGELOG.md

it seems that does not work?
See #143, #141

aliothe: Fixed, handles 0 - n issue numbers that are automagically translated to md links for the issue(s)

.github/workflows/release-and-update-changelog.yml
aliothe: In case we ever want to push to dev and skip ci

scripts/create-initial-changelog.sh
aliothe: The initial CHANGELOG.md is already in the PR.
I put this in here in case we ever need to run in again - it is not used by any action.

.github/workflows/release-and-update-changelog.yml
aliothe: If there were no changes to the changelog then we will skip adding it, doing the amend commit and force-with-lease.
This is extra caution in case the changelog was not changed in way.

.github/workflows/release-and-update-changelog.yml
aliothe: changed

.gitignore
aliothe: ha ha, I haven't used that for this no but yeah, I have played with the java code to see what it can do :)

aliothe:
If you want to test this, create a branch from feature/not-dev ( it is the same as feature/changelog except for the branch name for the actions).
Then squash merge that feature into feature/not-dev and see what happens.

Then you can also check out what the actions do.

scripts/create-initial-changelog.sh
aliothe: https://github.com/casualcore/casual-java/blob/feature/not-dev/CHANGELOG.md

Just tested it, it works as it should.
All * items from the PR remains.

.github/workflows/validate-pr-title.yml
aliothe: That things validates the PR title, that it conforms to convetional commit messages.
If we want to also create and output the change log entry, that would be something different that we should add then - not to the validation of the PR title though.

scripts/create-initial-changelog.sh
cklellie: https://github.com/casualcore/casual-java/blob/feature/not-dev/CHANGELOG.md#3247---2024-11-21

If you look here, you can see it seems to have added an extra one bullet for the message summary and removed all bar the first one for the description. Rather than maintain them as a list.

CHANGELOG.md
```diff
@@ -0,0 +1,447 @@
...
+casual-api-spi-impl - Module that wraps all our spi modules, currently only casual-json-provider. This is the module that other applications should depend on.
+
+## [3.2.48] - 2024-11-28
+- Bugfix/3.2/inbound discovery with custom service handlers (#130)
+- Utilise all available ServiceHandlers to check if services are available during startup of inbound in Discover mode.
+
+## [3.2.47] - 2024-11-21
+- Bugfix/3.2/configuration mode not honoured (#126)
+- Refactoring configuration functionality to ensure correctness of resulting configuration values.
+Simplifying configuration to flatter property based structure for ease of retrieval.
```
cklellie: This is still not right - stripping the `*` is meaning that you are removing the list items from my pr message.
When you create  a PR - the commit messages from the branch are listed by default with `*`, so we need to not break this.

.github/workflows/release-and-update-changelog.yml
cklellie: Are you saying that this changelog generation is ci? That's a bit of a stretch. Equally, I don't see why we would want to control it like this. It would mean that PR title would need to be "feat: my change [skip ci]", which seems off to me.

CHANGELOG.md
aliothe: "To create an unordered list, add dashes (-), asterisks (*), or plus signs (+) in front of line items. Indent one or more items to create a nested list.", https://www.markdownguide.org/basic-syntax/

Did we not discuss this yesterday already?

scripts/create-initial-changelog.sh
aliothe: Fixed

.github/workflows/validate-pr-title.yml
aliothe: Fixed

CHANGELOG.md
aliothe: Fixed

.github/workflows/release-and-update-changelog.yml
aliothe: Fixed

CHANGELOG.md
```diff
@@ -0,0 +1,521 @@
+# Changelog
+This is the changelog for *casual java* and all changes are listed in this document.
+
+
+## [3.3.1] - 2025-01-13
+### Feature/3.2/unmarshall all null values (#143)
```
cklellie: For consistency, we should add the links for these issues here too.

.github/workflows/validate-pr-title.yml
```diff
@@ -0,0 +1,24 @@
+name: Validate PR Title
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
```
cklellie: As we are only validating the title, `synchronize` can be removed from here. Reducing the github actions minutes usage.

scripts/utilities.py
```diff
@@ -0,0 +1,81 @@
...
+def replace_issue_numbers(commit_msg):
+    """ Function to replace issue numbers with markdown link
+    such as:
+    #132 -> [#132](https://github.com/casualcore/casual-java/issues/132)
+    """
+    pattern = r'#(\d+)'
+
+    def replacement(match):
+        number = match.group(1)
+        return f'[{match.group(0)}](https://github.com/casualcore/casual-java/issues/{number})'
```
cklellie: This can be retrieved from the event information.
`github.event.repository.issues_url`
The format of the url you receive seems to conform to RFC 6570 URI templating, so `https://github.com/casualcore/casual-java/issues{/number}`, which you should be able to format nicely.

The only slight wrinkle I see is that it needs to be slightly modified as that is the api link, rather than the html link.
We also recieve the api and server urls in the same event payload so can easily fix that too:
Just run a replace on the link with the value from:
`github.event.api_url`
with:
`github.event.server_url`

cklellie: If you want to see the full details.
```
      - name: Update changelog
        env:
          GITHUB_CONTEXT: ${{ toJson(github)}}
        run: |
          echo "Hi there, here is the context:"
          echo "$GITHUB_CONTEXT"
```

cklellie: Looks like a lot of the data is actually already available as envs under `GITHUB_`, i just ran with `env | grep GITHUB | sort` and saw `GITHUB_API_URL` and `GITHUB_SERVER_URL` are already there :)

cklellie (approved):

Issue #137, opened 2024-11-28
Maven Central Publish with Tokens


cklellie:
The existing publish.md and publish.gradle rely on the old username and password approach.
It is now required that the publish is performed using a token that needs to be generated for the users sonar type account instead.
Otherwise you receive a 401 HTTP response when attempting to publish.

Changing the secret values of `ossrhUsername` and `ossrhPassword` inside the gradle.properties to the token varient, works, though this is confusing when the naming does not change to reflect they are now tokens.

It should therefore be changed in the gradle.properties, publish.gradle file to use a new name of for example:
`ossrhTokenUser` 
`ossrhTokenPassword`

It should also be updated in the publish.md to explain how to generate this token.

cklellie:
updated as part of #184

Issue #132, opened 2024-11-28
Configuration enum values allow case insensitive


cklellie:
Currently if you set for example:
Inbound Startup Mode you must use lowercase.
Netty Logging you must use UPPERCASE.

It would be "nicer" if we just accepted the values in the configuration case insensitively.

Such that `INFO == info == InFo` or `DISCOVER == discover == DiscOVEr`.

This should be in both environment variables and through the json configuration file.

PR #173, opened 2025-03-27
feat: testing changelog changes

cklellie:
I thought the description would include all the commit messages - that used to happen.

* but i did not get it this time.
* Perhaps it comes next time.

What about adding an issue reference #137 because why not.
How about another reference #132


cklellie:
This comment will not cause the workflows to run.

PR #183, opened 2025-04-15
feat: inbound gw protocol bumps casual 1.6

aliothe:
casual-jca now supports protocol version 1.0, 1.1, 1.2 on both outbound, inbound and reverse inbound connections.

Previously inbound and reverse inbound were locked to 1.0.


casual/casual-api/src/test/groovy/se/laz/casual/network/ProtocolMatcherTest.groovy
```diff
@@ -0,0 +1,34 @@
...
+{
+   def 'null'()
+   {
+      when:
+      ProtocolMatcher.match(null)
+      then:
+      thrown(NullPointerException)
+   }
+
+   def 'correctness'()
```
cklellie, line 22: Add a few more tests:
* unsorted input.
* non sequential.
* induce the ProtocolVersionException - when there is no match.

casual/casual-api/src/main/java/se/laz/casual/network/ProtocolMatcher.java
```diff
@@ -0,0 +1,32 @@
...
+    {}
+    public static Long match(List<Long> protocolVersions)
+    {
+        Objects.requireNonNull(protocolVersions,"protocolVersions can not be null" );
+        List<Long> supportedVersions = ProtocolVersion.supportedVersionNumbers();
+        List<Long> matchedVersions = protocolVersions.stream().filter(supportedVersions::contains).sorted().collect(Collectors.toList());
+        Collections.sort(matchedVersions);
+        if(matchedVersions.isEmpty())
+        {
+            // note: this should never be able to happen since all combinations of casual/casual-jca at least supports version 1.0
```
cklellie, line 26: At the moment, yes, it should not happen, but nothing to stop it happening in the future if this code remains unchanges and casual decided to not support. So unlikely, but possible.

casual/casual-api/src/main/java/se/laz/casual/network/ProtocolVersionException.java
```diff
@@ -0,0 +1,20 @@
...
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+
+package se.laz.casual.network;
+
+import se.laz.casual.api.CasualRuntimeException;
+
+import java.util.function.Supplier;
+
+public class ProtocolVersionException extends CasualRuntimeException
```
cklellie: We should have this thrown in the ProtocolVersion unmarshall method now that you have created it.
Currently that throws CasualConnectionException.
If you change this to extend CasualConnectionException, that would be a simple drop in replacement and be more semantically correct.

casual/casual-network/src/main/java/se/laz/casual/network/InboundMsgSender.java
```diff
@@ -0,0 +1,56 @@
...
+import se.laz.casual.config.ConfigurationService;
+import se.laz.casual.network.protocol.messages.CasualNWMessageImpl;
+import se.laz.casual.network.protocol.messages.domain.DomainDisconnectRequestMessage;
+import se.laz.casual.network.protocol.messages.domain.DomainDiscoveryTopologyUpdateMessage;
+
+import java.util.Objects;
+import java.util.UUID;
+import java.util.logging.Logger;
+
+public class InboundMsgSender
```
cklellie: `InboundMessageSender` - just for consistency with everywhere else where we use Message.

casual/casual-network/src/test/groovy/se/laz/casual/network/InboundDeactivatedContextTest.groovy
```diff
@@ -0,0 +1,60 @@
...
+      given:
+      def idOne = Mock(ChannelId)
+      def futureOne = Mock(ChannelFuture)
+      def channelOne = Mock(Channel){
+         isWritable() >> true
+         id() >> idOne
+         closeFuture() >> futureOne
+         1 * writeAndFlush({ CasualNWMessage<DomainDisconnectRequestMessage> msg ->
+            Objects.requireNonNull(msg, 'msg can not be null')
+            if(! (msg.getMessage() instanceof DomainDisconnectRequestMessage)){
```
cklellie, line 29: Should we check the contents of the disconnect message is correct?

cklellie (commented):
Generally looking good, just a few small things.
Also i put a branch up `feature/protocol-version-refactor`, if you want to include that in here.

casual/casual-api/src/main/java/se/laz/casual/network/ProtocolMatcher.java
aliothe: Yep, that's why that code and comment is there

casual/casual-network/src/test/groovy/se/laz/casual/network/InboundDeactivatedContextTest.groovy
aliothe: What content should we check in the message?

cklellie: i misread the diff - thought there was more than just a random execution id.

cklellie (approved):

PR #185, opened 2025-05-06
feat: handle graceful shutdown

aliothe:
When there is a graceful shutdown we:
* drain the current in flight transactions
* return TPENOENT for new outbound requests

### DIFF
--- a/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/CasualInboundTransactionRegistry.java
+++ b/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/CasualInboundTransactionRegistry.java
@@ -30,12 +30,10 @@ public class CasualInboundTransactionRegistry
         Objects.requireNonNull(key, "key can not be null");
         Objects.requireNonNull(channelId, "channelId can not be null");
         log.finest(() -> "Removing transaction " + key + " from transaction registry");
-        Set<XidKey> ids = transactions.get(channelId);
-        ids.remove(key);
-        if(ids.isEmpty())
-        {
-            transactions.remove(channelId);
-        }
+        transactions.computeIfPresent(channelId, (id, set) ->{
+           set.remove(key);
+           return set.isEmpty() ? null : set;
+        });
     }
 
     public void remove(ChannelId channelId)
--- a/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/CasualInboundTransactionRegistry.java
+++ b/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/CasualInboundTransactionRegistry.java
@@ -17,18 +17,20 @@ import java.util.logging.Logger;
 public class CasualInboundTransactionRegistry
 {
     private static final Logger log = Logger.getLogger(CasualInboundTransactionRegistry.class.getName());
+    private static final String CHANNEL_ID_CAN_NOT_BE_NULL = "channelId can not be null";
     private final Map<ChannelId, Set<XidKey>> transactions = new ConcurrentHashMap<>();
 
     public void add(ChannelId channelId, XidKey key)
     {
+        Objects.requireNonNull(channelId, CHANNEL_ID_CAN_NOT_BE_NULL);
         Objects.requireNonNull(key, "key can not be null");
         log.finest(() -> "Adding transaction " + key + " to inbound transaction registry");
         transactions.computeIfAbsent(channelId, id -> ConcurrentHashMap.newKeySet()).add(key);
     }
     public void remove(ChannelId channelId, XidKey key)
     {
+        Objects.requireNonNull(channelId, CHANNEL_ID_CAN_NOT_BE_NULL);
         Objects.requireNonNull(key, "key can not be null");
-        Objects.requireNonNull(channelId, "channelId can not be null");
         log.finest(() -> "Removing transaction " + key + " from transaction registry");
         transactions.computeIfPresent(channelId, (id, set) ->{
            set.remove(key);
@@ -38,7 +40,7 @@ public class CasualInboundTransactionRegistry
 
     public void remove(ChannelId channelId)
     {
-        Objects.requireNonNull(channelId, "channelId can not be null");
+        Objects.requireNonNull(channelId, CHANNEL_ID_CAN_NOT_BE_NULL);
         log.finest(() -> "Removing all pending transactions for " + channelId + " from transaction registry");
         transactions.remove(channelId);
     }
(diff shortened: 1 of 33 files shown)


casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/CasualInboundTransactionRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+        Objects.requireNonNull(key, "key can not be null");
+        log.finest(() -> "adding transaction " + key + " for channel id: " + channelId + " to inbound transaction registry");
+        transactions.computeIfAbsent(channelId, id -> ConcurrentHashMap.newKeySet()).add(key);
+    }
+    public void remove(ChannelId channelId, XidKey key)
+    {
+        Objects.requireNonNull(channelId, CHANNEL_ID_CAN_NOT_BE_NULL);
+        Objects.requireNonNull(key, "key can not be null");
+        log.finest(() -> "removing transaction " + key + "for channel id: " + channelId +" from transaction registry");
+        transactions.computeIfPresent(channelId, (id, set) ->{
```
cklellie, line 35: I think we have a race condition here.

If add is called concurrently to remove, after the set.isEmpty check, something is added, but we have then placed null rather than the set back.
Could we prevent this by just leaving the empty set in place rather than removing, then if something is later added ot the channel it will just use the empty set to add the next one.
Yes a bit of memory overhead, but it will remove this race and also slightly reduce the add performance as we won't need to make a new set each time the channels transactions are all completed, which i think will likely be a common situations.
run a few transactions, idle, run a few more, etc.

casual/casual-inbound-api/src/test/groovy/se/laz/casual/jca/inflow/XidKeyTest.groovy
```diff
@@ -0,0 +1,36 @@
...
+      byte[] bqualData = (0..XID.MAX_XID_DATA_SIZE / 2 - 1) as byte[]
+      def formatTypeOne = 42L
+      def formatTypeTwo = 99L
+      when:
+      Xid xidOne = XID.of(gtridData, bqualData, formatTypeOne)
+      XidKey keyOne = XidKey.of(xidOne)
+      Xid xidTwo = XID.of(gtridData, bqualData, formatTypeTwo)
+      XidKey keyTwo = XidKey.of(xidTwo)
+      then:
+      keyOne == keyOne
```
cklellie, line 30: Add an additional key which is identical to keyOne to then check equality when they do not have referencial equality but are equal.

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
```diff
@@ -239,11 +247,11 @@ public void prepareRequest(CasualNWMessage<CasualTransactionResourcePrepareReque
     }
 
     @Override
-    public void commitRequest(CasualNWMessage<CasualTransactionResourceCommitRequestMessage> message, Channel channel, XATerminator xaTerminator)
+    public void commitRequest(CasualNWMessage<CasualTransactionResourceCommitRequestMessage> message, Channel channel, XATerminator xaTerminator, CasualInboundTransactionRegistry inboundTransactionRegistry)
     {
         log.finest(() -> "commitRequest(). " + PrettyPrinter.format(message.getCorrelationId(), message.getMessage().getExecution(), message.getMessage().getXid()) + message);
-
         Xid xid = message.getMessage().getXid();
+        inboundTransactionRegistry.remove(channel.id(), XidKey.of(xid));
```
cklellie: why removing so early? Should this not be lower down in the finally?

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
```diff
@@ -271,12 +279,12 @@ public void commitRequest(CasualNWMessage<CasualTransactionResourceCommitRequest
...
 
     @Override
-    public void requestRollback(CasualNWMessage<CasualTransactionResourceRollbackRequestMessage> message, Channel channel, XATerminator xaTerminator)
+    public void requestRollback(CasualNWMessage<CasualTransactionResourceRollbackRequestMessage> message, Channel channel, XATerminator xaTerminator, CasualInboundTransactionRegistry inboundTransactionRegistry)
     {
         log.finest(() -> "requestRollback(). " + PrettyPrinter.format(message.getCorrelationId(), message.getMessage().getExecution(), message.getMessage().getXid()) + message );
 
         Xid xid = message.getMessage().getXid();
-
+        inboundTransactionRegistry.remove(channel.id(), XidKey.of(xid));
```
cklellie: move down to finally?

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
```diff
@@ -133,6 +135,8 @@ public void endpointActivation(MessageEndpointFactory endpointFactory,
                                    ActivationSpec spec) throws ResourceException
     {
         log.info(()->"start endpointActivation() ");
+        inboundTransactionRegistry = new CasualInboundTransactionRegistry();
+        RuntimeInformation.setDomainIsBeingShutdown(false);
```
cklellie, line 139: why would this be true at this point?

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualResourceAdapter.java
```diff
@@ -219,10 +225,15 @@ private void startWork(Work work, WorkListener workListener)
     public void endpointDeactivation(MessageEndpointFactory endpointFactory,
                                      ActivationSpec spec)
     {
-        log.finest(()->"endpointDeactivation()");
+        log.info(()->"endpointDeactivation() ");
+        RuntimeInformation.setDomainIsBeingShutdown(true);
         InboundDeactivatedContext.domainDisconnect();
         InboundDeactivatedContext.clear();
         InboundTopologyUpdateContext.clear();
+        Predicate predicate = () -> inboundTransactionRegistry.hasPending() || CasualResourceManager.getInstance().hasPending();
```
cklellie, line 233: What is the difference here, would `inboundTransactionRegistry.hasPending()`, not always be a the same as `CasualResourceManager.getInstance().hasPending()`?

cklellie (commented):
What about when the shutdown is initiated and a new inbound request is received?
Should we not be shutting down the inbound server once the endpointDeactivation method is called?

cklellie (approved):
Approved for when you have fixed the additional equality test.

PR #189, opened 2025-06-05
refactor: Replace java.util.logging.Logger with System.Logger

MedTahiri:
Replaced java.util.logging.Logger with System.Logger to allow for more flexible and future-proof logging. Behavior remains the same.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfo.java
@@ -8,6 +8,8 @@ package se.laz.casual.api.buffer.type.fielded;
 
 import java.nio.charset.Charset;
 
+import static java.lang.System.Logger.Level.*;
+
 /**
  * Encoding information to use for string encoding/decoding
  * @see EncodingInfoProvider
@@ -22,7 +24,7 @@ public final class EncodingInfo
     {
         EncodingInfoProvider p = EncodingInfoProvider.of();
         charset = p.getCharset();
-        log.log(System.Logger.Level.INFO,() -> "casual fielded encoding set to charset: " + charset);
+        log.log(INFO,() -> "casual fielded encoding set to charset: " + charset);
     }
 
     /**
--- a/casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfo.java
@@ -7,7 +7,6 @@
 package se.laz.casual.api.buffer.type.fielded;
 
 import java.nio.charset.Charset;
-import java.util.logging.Logger;
 
 /**
  * Encoding information to use for string encoding/decoding
@@ -15,7 +14,7 @@ import java.util.logging.Logger;
  */
 public final class EncodingInfo
 {
-    private static final Logger log = Logger.getLogger(EncodingInfo.class.getName());
+    private static final System.Logger log = System.getLogger(EncodingInfo.class.getName());
     private static final Charset charset;
     private EncodingInfo()
     {}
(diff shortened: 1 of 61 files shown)


casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfo.java
```diff
@@ -7,23 +7,22 @@
...
+    private static final System.Logger log = System.getLogger(EncodingInfo.class.getName());
     private static final Charset charset;
     private EncodingInfo()
     {}
     static
     {
         EncodingInfoProvider p = EncodingInfoProvider.of();
         charset = p.getCharset();
-        log.info(() -> "casual fielded encoding set to charset: " + charset);
+        log.log(System.Logger.Level.INFO,() -> "casual fielded encoding set to charset: " + charset);
```
cklellie: Use static import for Level, please.

casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfoProvider.java
```diff
@@ -45,7 +43,7 @@ public Charset getCharset()
         }
         catch(IllegalArgumentException e)
         {
-            log.log(Level.WARNING, e, () -> "could not find charset by name: " + name + " falling back to utf-8");
+            log.log(System.Logger.Level.WARNING, () -> "could not find charset by name: " + name + " falling back to utf-8");
```
cklellie: Missing exception `e` in the log message. Please add back.

Also update all other locations this has happened.

casual/casual-event-client/src/main/java/se/laz/casual/event/client/EventClient.java
```diff
@@ -109,7 +108,7 @@ protected void initChannel(SocketChannel ch)
                         }
                     }
                 });
-        LOG.finest(() -> "about to connect to: " + clientInformation.getConnectionInformation().getAddress());
+        LOG.log(System.Logger.Level.TRACE,() -> "about to connect to: " + clientInformation.getConnectionInformation().getAddress());
```
cklellie: Please use `DEBUG` rather than `TRACE` to replace `finest()`.

This is one of the reason we are moving over to System.Logger so we can use DEBUG instead of finest.
There are a few places where TRACE is perhaps more appropriate, but set them all to DEBUG For now and we will review logging seperately in a later PR.

cklellie (changes requested):
Hello again! Thank you also for this PR.

As with Casual Caller, I have reviewed and added very similar comments for fixes and changes.

Again, please bump the version here too to the next patch version so we can merge.

Thank you :)

MedTahiri:
Hello,

I've fixed the issues as requested, similar to how it's handled in casual-caller.
Let me know if anything else needs to be adjusted.

Thanks!

casual/casual-api/src/main/java/se/laz/casual/api/buffer/type/fielded/EncodingInfo.java
```diff
@@ -7,23 +7,24 @@
 package se.laz.casual.api.buffer.type.fielded;
 
 import java.nio.charset.Charset;
-import java.util.logging.Logger;
+
+import static java.lang.System.Logger.Level.*;
```
cklellie, line 11: As with casual-caller - we have a preference for not using wildcard imports. Please update accordingly in all files.

casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
```diff
@@ -119,7 +120,7 @@ private void issueCall()
         {
             replyBuilder.setError( ErrorState.TPENOENT )
                         .setTransactionState( TransactionState.ROLLBACK_ONLY );
-            log.warning( ()-> "ServiceHandler not available for: " + message.getServiceName() );
+            log.log(WARNING,()-> "ServiceHandler not available for: " + message.getServiceName(),e );
```
cklellie, line 123: `e` is not needed here as the message is sufficient. Please remove.

cklellie (changes requested):
Thank you for the updates, looking good, just 2 small changes please.

PR #197, opened 2026-03-02
feat: Add support for casual gateway protocol versions 1.3 and 1.4

aliothe:
Adds full support for casual 1.7 and gateway protocol versions 1.3 and 1.4, introducing distributed tracing via span IDs, structured queue error codes and extended queue discovery metadata.                                                                                                                       

Compatibility tested with casual versions:
* 1.6.18
* 1.7.10
* `>=` 1.8.13 (earlier issues fixed in this release)

Protocol 1.3                                                                                                                                                                                
                                                                                                                                                                                              
  - Distributed tracing: Service call requests and conversation connect messages now carry a parentSpan (8-byte span ID), enabling end-to-end tracing across services. Tracing context        
  propagates across thread boundaries via InboundThreadLocal and the Concurrent utility.                                                                                                      
  - Queue error codes: Enqueue and dequeue replies carry a structured QueueErrorCode (OK, NO_MESSAGE, NO_QUEUE, ARGUMENT, SYSTEM, SIGNALED), surfaced via EnqueueReturn and DequeueReturn.    
  - Service call reply: XID removed from wire format for >= 1.3.                                                                                                                              
  - Dequeue reply: Limited to at most one message.
  
  Protocol 1.4

  - Extended queue discovery metadata: Discovery reply Queue objects gain retryDelay, enqueueEnabled, and dequeueEnabled fields, exposed via QueueDetails.
  
  For testing, we have changed the binary test files to base64 encoded files. This data can then be retrieved via the protocol description from casual. Since that data is only the actual payload without the header - a tool that creates complete base 64 encoded messages has been added.


casual/casual-api/src/main/java/se/laz/casual/api/concurrency/Concurrent.java
```diff
@@ -0,0 +1,73 @@
...
+
+package se.laz.casual.api.concurrency;
+
+import se.laz.casual.jca.InboundThreadContext;
+import se.laz.casual.jca.InboundThreadLocal;
+
+import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+
+public final class Concurrent
```
cklellie, line 20: Please add javadoc on the top of classes just to briefly explain its purpose - helps to then review what the code is doing inside and confirm if it is doing as described correctly.
Also i see no unit tests for this class?

casual/casual-jca/src/main/java/se/laz/casual/jca/service/OutboundContextCreator.java
```diff
@@ -0,0 +1,59 @@
...
+        return ProtocolVersion.isProtocolVersionGreaterOrEqualToOneThree(protocolVersion)
+                ? createForProtocolVersionThatSupportsSpanId(execution)
+                : createForProtocolThatDoesNotSupportSpanId(execution);
+    }
+
+    private static OutboundContext createForProtocolThatDoesNotSupportSpanId(UUID execution)
+    {
+        Optional<InboundThreadContext> inboundThreadContext = InboundThreadLocal.getContext();
+        // we still want to propagate the inbound execution and parentName if available ( call comes from inbound)
+        UUID effectiveExecution = inboundThreadContext.stream()
```
cklellie: Why stream and then findFirst/findAny as there will only ever be max one?
All these can be replaced with a simpler map and lambda e.g:
```java
 UUID effectiveExecution = inboundThreadContext.map( InboundThreadContext::execution ).orElse(execution);
 String parentName = inboundThreadContext.map( InboundThreadContext::parentName ).orElse("");
```

casual/casual-api/src/main/java/se/laz/casual/api/network/protocol/messages/CasualNWMessageType.java
```diff
@@ -21,20 +21,26 @@ public enum CasualNWMessageType
     DOMAIN_DISCONNECT_REPLY(7203),
     DOMAIN_DISCOVERY_REQUEST(7300),
     DOMAIN_DISCOVERY_REPLY(7301),
+    DOMAIN_DISCOVERY_REPLY_PROTOCOL_VERSION_EQUAL_OR_GREATER_TO_ONE_FOUR(7311),
```
cklellie: GREATER_TO? GREATER_THAN?
These names lengths are long which hinders readability.
Also what happens when a new version comes, say 1.6  - it will be wrong.
We should try and make the naming more definitive and factual so it will last over time.
e.g:
MESSAGE - if there is only one version. Need to maintain backwards compatability i guess?
MESSAGE_V_1_4 - for the version that introduces it.
So in this example we would have:
DOMAIN_DISCOVERY_REPLY_V_1_4
Then if a version 1.6 appears with a change too:
DOMAIN_DISCOVERY_REPLY_V_1_6

Then you have the following simple table for when you have the versions:
DOMAIN_DISCOVERY_REPLY -> 1.0, 1.1, 1.2, 1.3
DOMAIN_DISCOVERY_REPLY_V_1_4 -> 1.4, 1.5
DOMAIN_DISCOVERY_REPLY_V_1_6 -> 1.6 or higher until another change is made.

casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
```diff
@@ -15,12 +15,15 @@ public class DequeueReturn
 {
     private final QueueMessage queueMessage;
     private final ErrorState errorState;
+    // only available in protocol version >= 1.3
```
cklellie: Unnecessary comment, as you have it in the javadoc on the getter.

casual/casual-api/src/main/java/se/laz/casual/api/queue/EnqueueReturn.java
```diff
@@ -48,9 +59,11 @@ public int hashCode()
     @Override
     public String toString()
```
cklellie, line 73: Add updates to equals and hashcode for new field.

casual/casual-api/src/main/java/se/laz/casual/api/queue/QueueDetails.java
```diff
@@ -1,27 +1,37 @@
...
+import java.util.Optional;
 
 public class QueueDetails
 {
     private final String name;
     private final long retries;
-
-    public QueueDetails(String name, long retries)
+    private final ProtocolVersion protocolVersion;
+    // these are only available in protocol version >= 1.4
```
cklellie: Unnecessary comment as you have it in the getters.

casual/casual-api/src/main/java/se/laz/casual/api/queue/QueueDetails.java
```diff
@@ -1,27 +1,37 @@
...
+    private final Long retryDelay;
+    private final Boolean enqueueEnabled;
+    private final Boolean dequeueEnabled;
+    private QueueDetails(String name, long retries, ProtocolVersion protocolVersion, Long retryDelay, Boolean enqueueEnabled, Boolean dequeueEnabled)
```
cklellie: Just a single constructor with the Builder where you set the fields from the builder would be nicer here.

casual/casual-api/src/main/java/se/laz/casual/api/queue/QueueDetails.java
```diff
@@ -34,13 +44,126 @@ public long getRetries()
         return retries;
     }
 
+    /**
+     * Only available in gw protocol version >= 1.4
```
cklellie: There is an `@since` for javadoc which could be nicer to use for all these types of comments.
e.g. `@since gw protocol version 1.4.`
Then you can have normal javadoc before that.

casual/casual-api/src/main/java/se/laz/casual/api/queue/QueueDetails.java
```diff
@@ -34,13 +44,126 @@ public long getRetries()
...
+    }
+
     @Override
     public String toString()
     {
-        return "QueueDetails{" +
-                "name='" + name + '\'' +
-                ", retries=" + retries +
-                '}';
+        final StringBuilder sb = new StringBuilder("QueueDetails{");
```
cklellie: Do we really want to maintain `toString` like this - could we just have them auto generate? It's only for debug purposes that we have them. Technically the object now regardless of the protocol version does contain the additional fields, so might be nice to see they are "empty" when debugging quickly.

casual/casual-api/src/main/java/se/laz/casual/api/queue/QueueDetails.java
```diff
@@ -34,13 +44,126 @@ public long getRetries()
...
+
+    public static Builder createBuilder()
+    {
+        return new Builder();
     }
 
+    public static final class Builder
+    {
+        private String name;
+        private long retries;
```
cklellie, line 111: Previously, without builder, primitive `long` of `retries` was always explicitly provided. Now it can be implicitly defaulted to `0` - is that correct?

casual/casual-api/src/main/java/se/laz/casual/api/queue/QueueErrorCode.java
```diff
@@ -0,0 +1,42 @@
...
+    {
+        this.value = value;
+    }
+
+    public int getValue()
+    {
+        return value;
+    }
+
+    public static final QueueErrorCode unmarshal(int id)
```
cklellie, line 30: Intellij warns about final here. 
Final can be useful to stop children of classes from hiding statics. But this is an `enum` which can't be extended, so static and final for these marshall and unmarshall methods here is redundant hence the warning.
I see this pattern in a few other enums too, though some are not final - nice to be consistent :D
What do you want to do with these?

casual/casual-api/src/main/java/se/laz/casual/jca/SpanId.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public static SpanId of()
+    {
+        byte[] bytes = new byte[ID_LENGTH];
+        RANDOM.nextBytes(bytes);
+        return new SpanId(bytes);
+    }
+
+    public static SpanId of(long value)
```
cklellie: javadoc please.

casual/casual-api/src/main/java/se/laz/casual/network/ProtocolVersion.java
```diff
@@ -94,4 +96,29 @@ public static List<String> supportedVersions()
...
+    {
+        return protocolVersion.version >= ProtocolVersion.VERSION_1_2.version;
+    }
+
+    public static boolean supportsDomainDisconnect(ProtocolVersion protocolVersion)
+    {
+         return protocolVersion.version >= ProtocolVersion.VERSION_1_1.version;
+    }
+
+    public static boolean isProtocolVersionGreaterOrEqualToOneThree(ProtocolVersion protocolVersion)
```
cklellie: When reading these in the code it is difficult to read, as the version becomes text rather than numerical.

Can we perhaps have this as a non static methods ?

-  `> GT / isGreaterThan`
- `>= GTE / isGreaterThanOrEqualTo`
-  `< LT / isLessThan`
-  `<= LTE / isLessThanOrEqualTo`

so `version.isLTE( otherVersion )` or `version.isLessThatOrEqualTo( otherVersion )` etc

If not I also wonder also if these could be moved to another location like ProtocolMatcher or another new thing.
Similar with the previous `supportsX` methods, naming is good, but does it live here?

casual/casual-api/src/test/groovy/se/laz/casual/api/queue/EnqueueReturnTest.groovy
```diff
@@ -36,22 +49,28 @@ class EnqueueReturnTest extends Specification {
...
 
-        EnqueueReturn enqueueReturn = EnqueueReturn.createBuilder().withId(uuid).withErrorState(errorState).build()
+        EnqueueReturn enqueueReturn = EnqueueReturn.createBuilder()
+                .withId(uuid)
+                .withErrorState(errorState)
+                .withErrorCode(errorCode)
+                .build()
 
         then:
         enqueueReturn != null
```
cklellie, line 90: ok, that explains why the equals and hashcode update missing was not caught in test.
Need to do more than check it is not null!

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+        }
+    }
+
+    def "of() should reject null context with meaningful NPE"()
+    {
+        when:
+        InboundThreadLocal.of(null)
+
+        then:
+        NullPointerException e = thrown()
```
cklellie: just `thrown NullPointerException` then.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+    InboundThreadContext testContext
+    InboundThreadContext altContext
+    SpanId testSpanId
+    SpanId altSpanId
+    UUID testExecution
+    UUID altExecution
+
+    def setup()
+    {
+        testSpanId = SpanId.of(42L)
```
cklellie: Why not shared?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+    {
+        when:
+        def inboundThreadLocal = InboundThreadLocal.of(testContext)
+        
+        then:
+        InboundThreadLocal.getContext().isPresent()
+        InboundThreadLocal.getContext().get() == testContext
+        
+        cleanup:
+        inboundThreadLocal?.close()
```
cklellie: unnecessary duplicate cleanup from setup?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+    }
+
+    def "getContext() should return empty Optional when no context is set"()
+    {
+        expect:
+        !InboundThreadLocal.getContext().isPresent()
```
cklellie: Negation in assertion which is not following the descriptions - yes technically correct, but a simple `isEmpty` would be more readable.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+    {
+        given:
+        def inboundThreadLocal = InboundThreadLocal.of(testContext)
+        assert InboundThreadLocal.getContext().isPresent()
+
+        when:
+        inboundThreadLocal.close()
+
+        then:
+        !InboundThreadLocal.getContext().isPresent()
```
cklellie: `isEmpty`?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+        def inboundThreadLocal = InboundThreadLocal.of(testContext)
+
+        when:
+        inboundThreadLocal.close()
+        inboundThreadLocal.close()
+        inboundThreadLocal.close()
+
+        then:
+        !InboundThreadLocal.getContext().isPresent()
+        noExceptionThrown()
```
cklellie: unnecessary noExceptionThrown - only needed if there is nothing to assert upon other than that there is no excepiton. If there is an exception it would fail per default.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+    def "try-with-resources should automatically clean up context"()
+    {
+        when:
+        Optional<InboundThreadContext> contextAfterTry
+        try (def ignored = InboundThreadLocal.of(testContext))
+        {
+            assert InboundThreadLocal.getContext().isPresent()
+            assert InboundThreadLocal.getContext().get() == testContext
+        }
+        contextAfterTry = InboundThreadLocal.getContext()
```
cklellie: why the split assignment?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,397 @@
...
+       try (def ignored = InboundThreadLocal.of(testContext)) {
+          def latch = new CountDownLatch(1)
+
+          def work = {
+             childContext = InboundThreadLocal.getContext().orElse(null)
+             latch.countDown()
+          }
+          def wrapped = Concurrent.wrap(work)
+
+          new Thread(wrapped).start()
```
cklellie: I understand you want to check difference between a thread created as a child, compared with an executor which is not connected - but that is related to when you used InheritedThreadLocal I guess? 
I don't like Threads which can be left running forever in tests. 
Can this all be wrapped into an executor? But then you have an executor starting a thread - so perhaps it's the same end result of a thread left running forever.
Do we still need this test?

casual/casual-api/src/main/java/se/laz/casual/jca/InboundThreadLocal.java
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright (c) 2025 - 2026, The casual project. All rights reserved.
+ *
+ * This software is licensed under the MIT license, https://opensource.org/licenses/MIT
+ */
+package se.laz.casual.jca;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class InboundThreadLocal implements AutoCloseable
```
cklellie: javadoc explanation of class purpose please as name is not clear what this does.
Is it a link in a chain of parent to child thread contexts - or will it only be 1 parent and a child or?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+       childContext == testContext
+
+       when:
+       childContext = InboundThreadLocal.getContext().orElse(null)
+       then:
+       childContext == null
+    }
+
+    @Unroll
+    def "context validation - all required fields should be non-null #spanId #parentName #execution"()
```
cklellie: What is the purpose of this test?
It should not live here if you really want to test a java record then you should do that in an `InboundThreadContextTest.groovy`

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundContextScopeTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+        where:
+        spanId         | parentName    | execution
+        SpanId.of(1L)  | "test"        | UUID.randomUUID()
+        SpanId.of(999) | "validName"   | UUID.randomUUID()
+    }
+
+    def 'does not propagate to pooled thread if not wrapped - using Runnable'()
+    {
+       given:
+       def capturedInAsync = new AtomicReference<InboundThreadContext>()
```
cklellie, line 119: Why change to AtomicReference from the countdown latch approach you had previously?
AtomicReference initialises to null, so it could just mean that nothing happened, rather than validating the state after the executor has completed.
With a latch you will know that it failed to execute within the time allowed, with this approach you do not.
Or like you have later on, use the future returned from the submit with `get` timeout which throws if timeout is exceeded.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+          }
+          def wrapped = Concurrent.wrap(work)
+
+          new Thread(wrapped).start()
+
+          latch.await(5, TimeUnit.SECONDS)
+       }
+
+       then:
+       childContext == testContext
```
cklellie: There is no confirmation that the latch has counted down. You need to check the return from the await or check that the count is now zero here.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+        SpanId.of(999) | "validName"   | UUID.randomUUID()
+    }
+
+    def 'does not propagate to pooled thread if not wrapped - using Runnable'()
+    {
+       given:
+       def capturedInAsync = new AtomicReference<InboundThreadContext>()
+       when:
+       try (def ignored = InboundThreadLocal.of(testContext)) {
+          def executor = Executors.newFixedThreadPool(2)
```
cklellie: why pool size 2? works with 1, so you could also just use `newSingleThreadExecutor`.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+   def 'does propagate to pooled thread when wrapped - using Runnable'()
+   {
+      given:
+      def capturedInAsync = new AtomicReference<InboundThreadContext>()
+      when:
+      try (def ignored = InboundThreadLocal.of(testContext))
+      {
+         def executor = Executors.newFixedThreadPool(2)
```
cklellie: `newSingleThreadExecutor`

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundContextScopeTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+         executor.submit(Concurrent.wrap(work))
+         executor.shutdown()
+         executor.awaitTermination(5, TimeUnit.SECONDS)
+      }
+      then:
+      capturedInAsync.get() == testContext
+      InboundThreadLocal.getContext().isEmpty()
+   }
+
+   def 'does propagate to pooled thread when wrapped - using Callable'()
```
cklellie, line 154: Yet another approach for the same test but with a Callable.
Here you are using futures from the return of the submit to wait and then wait again.
Please pick an approach and stick to it so we have some consistency.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundContextScopeTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+         }
+      }
+      Callable<InboundThreadContext> wrappedCallable = Concurrent.wrap(work)
+      def future = executor.submit(wrappedCallable)
+      def result = future.get(5, TimeUnit.SECONDS)
+      executor.shutdown()
+      executor.awaitTermination(5, TimeUnit.SECONDS)
+
+      then:
+      captured.get() == null
```
cklellie, line 205: Yet another approach - why the double check - how will the return not be the same as the capture?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+   }
+
+   def 'wrapped Supplier should propagate context to pooled thread (CompletableFuture style)'() {
+      given:
+      def capturedInAsync = new AtomicReference<InboundThreadContext>()
+
+      when:
+      def result
+      try (def ignored = InboundThreadLocal.of(testContext)) {
+         def executor = Executors.newFixedThreadPool(2)
```
cklellie: `newSingleThreadExecutor`

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundContextScopeTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+
+         def future = CompletableFuture.supplyAsync(wrappedSupplier, executor)
+         result = future.get(5, TimeUnit.SECONDS)
+
+         executor.shutdown()
+         executor.awaitTermination(5, TimeUnit.SECONDS)
+      }
+
+      then:
+      capturedInAsync.get() == testContext
```
cklellie, line 232: Why capture and future return?

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundThreadLocalTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+         executor.awaitTermination(5, TimeUnit.SECONDS)
+      }
+      then:
+      captured1.get() == testContext
+      captured2.get() == testContext
+      captured3.get() == testContext
+      InboundThreadLocal.getContext().isEmpty()
+   }
+
+   def 'double close is safe and idempotent'()
```
cklellie: Duplicate of `def "close() should be idempotent - multiple calls should be safe"()` ?

casual/casual-api/src/test/groovy/se/laz/casual/network/ProtocolVersionTest.groovy
```diff
@@ -51,4 +51,78 @@ class ProtocolVersionTest extends Specification
       ]
    }
 
+   def 'isProtocolVersionGreaterOrEqualToOneTwo true'()
```
cklellie: Why seperate tests for true and false, just have a table with both `protocolVersion` and `expected` like you have further down.

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/CasualMessageListenerImpl.java
```diff
@@ -156,18 +160,19 @@ public void serviceCallRequest(CasualNWMessage<CasualServiceCallRequestMessage>
...
+        SpanId spanId = SpanId.of();
+        CasualServiceCallWork work = new CasualServiceCallWork(message.getCorrelationId(), message.getMessage(), isTpNoReply, protocolVersion, spanId);
 
         try
         {
             if(!isTpNoReply && isServiceCallTransactional( xid ) )
             {
                 inboundTransactionRegistry.add(channel.id(), XidKey.of(xid));
-                workManager.scheduleWork(work, WorkManager.INDEFINITE, createTransactionContext(xid, message.getMessage().getTimeout()), new ServiceCallWorkListener(channel, message.getMessage()));
+                workManager.scheduleWork(work, WorkManager.INDEFINITE, createTransactionContext(xid, message.getMessage().getTimeout()), new ServiceCallWorkListener(channel, message.getMessage(), spanId, protocolVersion));
```
cklellie, line 171: SpanId  and ProtocolVersion are both available from the work object which is available to `ServiceCallWorkListener`, so no need pass them in here again.

casual/casual-api/src/test/groovy/se/laz/casual/jca/InboundContextScopeTest.groovy
```diff
@@ -0,0 +1,346 @@
...
+      when:
+      def scope = InboundThreadLocal.of(testContext)
+      scope.close()
+      scope.close()
+
+      then:
+      noExceptionThrown()
+      InboundThreadLocal.getContext().isEmpty()
+   }
+}
```
cklellie, line 319: I would like to see a test where you simulation multiple inbound requests each calling outbounds on a shared executor to ensure that each child outbound gets it's correct inbound context.

casual/casual-jca/src/main/java/se/laz/casual/jca/discovery/CasualDiscoveryCaller.java
```diff
@@ -77,9 +78,22 @@ private DiscoveryReturn toDiscoveryReturn(CasualDomainDiscoveryReplyMessage mess
         return builder.build();
     }
 
-    private QueueDetails toQueueDetails(Queue queue)
+    private QueueDetails toQueueDetails(Queue queue, ProtocolVersion protocolVersion)
     {
-        return QueueDetails.of(queue.getName(), queue.getRetries());
+        return ProtocolVersion.isProtocolVersionGreaterOrEqualToOneFour(protocolVersion)
```
cklellie: Nicer to group common builder initialisation and then conditionally add additional fields based on protocol version.

casual/casual-jca/src/main/java/se/laz/casual/jca/queue/CasualQueueCaller.java
```diff
@@ -56,7 +60,17 @@ public EnqueueReturn enqueue(QueueInfo qinfo, QueueMessage msg)
         {
             // Always setting error state OK for now. In the future when error state is handled in the casual queue
             // protocol any error state supplied from casual should be used (same with dequeue)
-            return EnqueueReturn.createBuilder().withErrorState(ErrorState.OK).withId(makeEnqueueCall(UUID.randomUUID(), qinfo, msg)).build();
+            // This future thing should be removed in some release, however it means a new major release
+            // since it will break our already released API
+            CasualEnqueueReplyMessage replyMessage = makeEnqueueCall(UUID.randomUUID(), qinfo, msg);
+            EnqueueReturn.Builder builder = EnqueueReturn.createBuilder();
+            builder.withErrorState(ErrorState.OK)
```
cklellie, line 68: Why not have `ErrorState.OK` as default in the builder?

casual/casual-jca/src/main/java/se/laz/casual/jca/queue/CasualQueueCaller.java
```diff
@@ -71,7 +85,14 @@ public DequeueReturn dequeue(QueueInfo qinfo, MessageSelector selector)
         {
             // Always setting error state OK for now. In the future when error state is handled in the casual queue
             // protocol any error state supplied from casual should be used (same with enqueue)
-            return DequeueReturn.createBuilder().withErrorState(ErrorState.OK).withQueueMessage(makeDequeueCall(UUID.randomUUID(), qinfo, selector)).build();
+            // This future thing should be removed in some release, however it means a new major release
+            // since it will break our already released API
+            Pair<Optional<QueueMessage>, Optional<QueueErrorCode>> answer = makeDequeueCall(UUID.randomUUID(), qinfo, selector);
```
cklellie, line 91: Why are you doing this Pair stuff? Why not use same approach as `makeEnqueueCall` but returning `CasualDequeueReplyMessage`?
Then do the transform for messages and check protocol version and pull off message the error code? No need to create additional optionals.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/discovery/CasualDiscoveryCallerTest.groovy
```diff
@@ -0,0 +1,341 @@
...
+            getDomainId() >> domainOne
+            getProtocolVersion() >> protocolVersion
+        }
+        connection.networkConnection = networkConnection
+
+        when:
+        DiscoveryReturn result = instance.discover(corrid, serviceNames, queueNames)
+
+        then:
+        1 * networkConnection.request(_) >> {
```
cklellie, line 237: Add validation of request sent. as with previous tests.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/discovery/CasualDiscoveryCallerTest.groovy
```diff
@@ -0,0 +1,341 @@
...
+        def services = asServices([
+                [name: 'svc1', category: 'cat1', transactionType: TransactionType.AUTOMATIC, timeout: 0L, hops: 0L]
+        ])
+        def reply = createDiscoveryReply(services, [], ProtocolVersion.VERSION_1_2)
+
+        when:
+        DiscoveryReturn result = instance.discover(corrid, ['svc1'], [])
+
+        then:
+        1 * networkConnection.request(_) >> {
```
cklellie, line 291: Add validation of request message as with others.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/discovery/CasualDiscoveryCallerTest.groovy
```diff
@@ -0,0 +1,341 @@
...
+    {
+        setup:
+        def queues = asQueues(['q1'], ProtocolVersion.VERSION_1_2)
+        def reply = createDiscoveryReply([], queues, ProtocolVersion.VERSION_1_2)
+
+        when:
+        DiscoveryReturn result = instance.discover(corrid, [], ['q1'])
+
+        then:
+        1 * networkConnection.request(_) >> {
```
cklellie, line 318: Add validation of request message.

casual/casual-jca/src/test/groovy/se/laz/casual/jca/discovery/CasualDiscoveryCallerTest.groovy
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    def 'discover - network request fails, exception propagates'()
+    {
+        when:
+        instance.discover(corrid, serviceNames, queueNames)
+
+        then:
+        thrown(Exception)
+        1 * networkConnection.request(_) >> {
```
cklellie: This should be set in a `given:` block

casual/casual-jca/src/test/groovy/se/laz/casual/jca/discovery/CasualDiscoveryCallerTest.groovy
```diff
@@ -0,0 +1,341 @@
...
+    def 'discover - corrid is propagated to the network message'()
+    {
+        setup:
+        def reply = createDiscoveryReply([], [], ProtocolVersion.VERSION_1_2)
+
+        when:
+        instance.discover(corrid, serviceNames, queueNames)
+
+        then:
+        1 * networkConnection.request(_) >> {
```
cklellie: This should be set in a `given:` block.

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/decoding/decoders/conversation/ConnectRequestMessageDecoder.java
```diff
@@ -63,8 +69,25 @@ private static ConnectRequest createMessage(final byte[] data)
...
+        {
+            byte value = ByteBuffer.wrap(data, currentOffset, ConversationConnectRequestSizes.HAS_VALUE.getNetworkSize()).get();
+            currentOffset += ConversationConnectRequestSizes.HAS_VALUE.getNetworkSize();
+            hasTimeout = (value > 0);
+        }
+        long timeout = 0;
+        if(hasTimeout)
+        {
+            timeout = ByteBuffer.wrap(data, currentOffset, ConversationConnectRequestSizes.SERVICE_TIMEOUT.getNetworkSize()).getLong();
+            currentOffset += ConversationConnectRequestSizes.SERVICE_TIMEOUT.getNetworkSize();
```
cklellie, line 85: So `deadline.remaining` is the same size as `service.timeout.duration` and we are not exposing the change in naming here, even when we read?

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/decoding/decoders/service/CasualServiceCallRequestMessageDecoder.java
```diff
@@ -73,21 +78,46 @@ public CasualServiceCallRequestMessage readChunked(final ReadableByteChannel cha
...
+        long serviceTimeout = 0;
+        boolean hasTimeout = true;
+        if (ProtocolVersion.isProtocolVersionGreaterOrEqualToOneThree(protocolVersion))
+        {
+            byte value = ByteUtils.readFully(channel, ServiceCallRequestSizes.HAS_VALUE.getNetworkSize()).get();
+            hasTimeout = (value > 0);
+        }
+        if (hasTimeout)
+        {
+            serviceTimeout = ByteUtils.readFully(channel, ServiceCallRequestSizes.SERVICE_TIMEOUT.getNetworkSize()).getLong();
```
cklellie, line 93: Again are we not reflecting the naming change from timeout to deadline.remaining?

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/decoding/MessageVerifier.java
```diff
@@ -0,0 +1,137 @@
...
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.DOMAIN_DISCOVERY_REPLY_PROTOCOL_VERSION_EQUAL_OR_GREATER_TO_ONE_FOUR;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.DOMAIN_DISCOVERY_TOPOLOGY_UPDATE;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.ENQUEUE_REPLY;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.ENQUEUE_REPLY_PROTOCOL_VERSION_EQUAL_OR_GREATER_TO_ONE_THREE;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.SERVICE_CALL_REPLY;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.SERVICE_CALL_REPLY_PROTOCOL_VERSION_EQUAL_OR_GREATER_TO_ONE_THREE;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.SERVICE_CALL_REQUEST;
+import static se.laz.casual.api.network.protocol.messages.CasualNWMessageType.SERVICE_CALL_REQUEST_PROTOCOL_VERSION_EQUAL_OR_GREATER_TO_ONE_THREE;
+
+// it is not brain overload
```
cklellie: I agree with Sonar here - this could be simplified to a Map lookup check.
I have quickly refactored this and it worked, nice when you have so many tests available to verify!

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/messages/domain/Queue.java
```diff
@@ -48,6 +55,51 @@ public Queue setRetries(long retries)
...
+    public long getRetryDelay()
+    {
+        if(!ProtocolVersion.isProtocolVersionGreaterOrEqualToOneFour(protocolVersion))
+        {
+            throw new CasualProtocolException("retry delay not available in protocol version " + protocolVersion);
+        }
+        return retryDelay;
+    }
+
+    public Queue setRetryDelay(long retryDelay)
```
cklellie: Why setters? Can we not keep this immutable and move this over to a builder approach now it has more fields?

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/messages/domain/Queue.java
```diff
@@ -48,6 +55,51 @@ public Queue setRetries(long retries)
         return this;
     }
 
+    public long getRetryDelay()
+    {
+        if(!ProtocolVersion.isProtocolVersionGreaterOrEqualToOneFour(protocolVersion))
```
cklellie: Elsewhere we have had things optional, e.g. QueueDetails - but this is throwing. 
What did we decide in the end?

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/messages/queue/CasualDequeueReplyMessage.java
```diff
@@ -76,6 +111,10 @@ public String toString()
         final StringBuilder sb = new StringBuilder("CasualDequeueReplyMessage{");
         sb.append("execution=").append(execution);
         sb.append(", messages=").append(messages);
+        if(ProtocolVersion.isProtocolVersionGreaterOrEqualToOneThree(protocolVersion))
```
cklellie: again question about how we maintain toString implementations.

casual/casual-network-protocol/src/test/groovy/se/laz/casual/network/test/network/frombinary/CasualTransactionResourcePrepareReplyMessageTest.groovy
```diff
@@ -19,23 +20,19 @@ import spock.lang.Specification
...
 import java.nio.ByteBuffer
 
-/**
- * Created by aleph on 2017-04-03.
- */
 class CasualTransactionResourcePrepareReplyMessageTest extends Specification
 {
     @Shared
-    def resource = '/protocol/bin/message.transaction.resource.prepare.Reply.1000.5202.bin'
+    def resource = '/protocol/b64/message.transaction.resource.prepare.reply.1000.5202.b64'
```
cklellie, line 26: Please add documentation somewhere about how you generate these b64 files, so we have it available for the future.

casual/casual-network/src/main/java/se/laz/casual/network/inbound/ProtocolVersionValueHolder.java
```diff
@@ -0,0 +1,77 @@
...
+            // Thus, this should NEVER EVER HAPPEN
+            throw new CasualProtocolException("fatal: protocol version not set - should never be used in a context where it has not already been set");
+        }
+        return protocolVersion;
+    }
+
+    @Override
+    public void accept(ProtocolVersion protocolVersion)
+    {
+        Objects.requireNonNull(protocolVersion, "protocolVersion can not be null");
```
cklellie, line 53: Do we want to allow multiple updates to the version?

casual/casual-network/src/main/java/se/laz/casual/network/inbound/ProtocolVersionValueHolder.java
```diff
@@ -0,0 +1,77 @@
...
+package se.laz.casual.network.inbound;
+
+import se.laz.casual.api.network.protocol.messages.exception.CasualProtocolException;
+import se.laz.casual.network.ProtocolVersion;
+
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+public class ProtocolVersionValueHolder implements Supplier<ProtocolVersion>, Consumer<ProtocolVersion>
```
cklellie, line 16: I don't see the need for Supplier Consumer here. It would be more readable as a simple POJO.

casual/casual-network/src/main/java/se/laz/casual/network/inbound/ProtocolVersionValueHolder.java
```diff
@@ -0,0 +1,77 @@
...
+{
+    private volatile ProtocolVersion protocolVersion;
```
cklellie, line 26: Add an explanation of where/why we need `volatile` here.

concurrency.md
```diff
@@ -0,0 +1,93 @@
+# Concurrency and Inbound Context Propagation
+
+When an inbound service call is received, casual-java automatically establishes a thread-local context
```
cklellie, line 5: `Since/With protocol/version X...`

concurrency.md
```diff
@@ -0,0 +1,93 @@
+# Concurrency and Inbound Context Propagation
```
cklellie, line 1: Naming - yes this is about concurrency - but it's more about tracing.

casual/casual-api/src/main/java/se/laz/casual/api/queue/DequeueReturn.java
```diff
@@ -33,6 +36,15 @@ public ErrorState getErrorState()
         return errorState;
     }
 
+    /**
+     * Only available when using gw protocol version >= 1.3
```
cklellie: Please use `@since`

casual/casual-network-protocol/src/main/java/se/laz/casual/network/protocol/messages/domain/Queue.java
cklellie: API optional - internal - slap ourselves hard!

cklellie (commented):
As discussed, I will also push up an addition test for InboundThreadLocalTest and updated version of MessageVerifier that I showed you.

casual/casual-api/src/main/java/se/laz/casual/api/network/protocol/messages/CasualNWMessageType.java
aliothe: went with:
DOMAIN_DISCOVERY_REPLY_FROM_ONE_FOUR etc

TobiasLeo:
I would like to see some documentation about considerations about integrating tracing with other products (including the application server e.g. WildFly). What would be a good way to integrate span/parent with calls initated through casual inbound? Whats needs to be done when calls are started outside of casual (for instance through a HTTP inbound call or calls that originate from a domain that then need to perform casual outbound calls where you want tracing to work across from non-casual into the realm of casual? Integration is much easier if a vendor can point to best practices

aliothe:
> I would like to see some documentation about considerations about integrating tracing with other products (including the application server e.g. WildFly). What would be a good way to integrate span/parent with calls initated through casual inbound? Whats needs to be done when calls are started outside of casual (for instance through a HTTP inbound call or calls that originate from a domain that then need to perform casual outbound calls where you want tracing to work across from non-casual into the realm of casual? Integration is much easier if a vendor can point to best practices

That's quite some undertaking perhaps, no idea so would need to spend quite some time to investigate and perhaps even experiment with what actually works - should we do that in this feature???
This feature is to be feature complete with casual 1.7 and protocol version up to 1.4.

Maybe a POC to do what you suggest should be its own feature - if you do not already know that it is tiny, then that work can go into here as well?

TobiasLeo:
> > I would like to see some documentation about considerations about integrating tracing with other products (including the application server e.g. WildFly). What would be a good way to integrate span/parent with calls initated through casual inbound? Whats needs to be done when calls are started outside of casual (for instance through a HTTP inbound call or calls that originate from a domain that then need to perform casual outbound calls where you want tracing to work across from non-casual into the realm of casual? Integration is much easier if a vendor can point to best practices
> 
> That's quite some undertaking perhaps, no idea so would need to spend quite some time to investigate and perhaps even experiment with what actually works - should we do that in this feature??? This feature is to be feature complete with casual 1.7 and protocol version up to 1.4.
> 
> Maybe a POC to do what you suggest should be its own feature - if you do not already know that it is tiny, then that work can go into here as well?

I'm really not looking at anything super in depth here or a complete solution. Just quick pointers to stuff like (just giving examples, don't know if they would be considered correct/viable within the solution of tracing in casual-jca)
- Want to continue the trace outside of casual for an inbound call? Adopt the span/parent from `InboundContextScope::getContext`
- Got a call from outside of casual and want to continue a trace with casual? Create a context in `InboundContextScope` and add your context values to ensure outbound calls use your context, you are required to clear the context you create
- Span/parent id - what values to use for spanId, parentName and execution

But maybe I'm out of scope for this PR and that sort of work would be a separate task that looks deeper into the aforementioned

aliothe:
> > > I would like to see some documentation about considerations about integrating tracing with other products (including the application server e.g. WildFly). What would be a good way to integrate span/parent with calls initated through casual inbound? Whats needs to be done when calls are started outside of casual (for instance through a HTTP inbound call or calls that originate from a domain that then need to perform casual outbound calls where you want tracing to work across from non-casual into the realm of casual? Integration is much easier if a vendor can point to best practices
> > 
> > 
> > That's quite some undertaking perhaps, no idea so would need to spend quite some time to investigate and perhaps even experiment with what actually works - should we do that in this feature??? This feature is to be feature complete with casual 1.7 and protocol version up to 1.4.
> > Maybe a POC to do what you suggest should be its own feature - if you do not already know that it is tiny, then that work can go into here as well?
> 
> I'm really not looking at anything super in depth here or a complete solution. Just quick pointers to stuff like (just giving examples, don't know if they would be considered correct/viable within the solution of tracing in casual-jca)
> 
> * Want to continue the trace outside of casual for an inbound call? Adopt the span/parent from `InboundContextScope::getContext`
> * Got a call from outside of casual and want to continue a trace with casual? Create a context in `InboundContextScope` and add your context values to ensure outbound calls use your context, you are required to clear the context you create
> * Span/parent id - what values to use for spanId, parentName and execution
> 
> But maybe I'm out of scope for this PR and that sort of work would be a separate task that looks deeper into the aforementioned

My feeling is that it is out of scope for this but that it should surely be done.
Maybe you want to create an issue and perhaps even look at it at some point as I know you've already done quite a bit of work work with open telemetry and wildfly?
That would then be complete with actual working examples how to integrate it.

TobiasLeo:
> > > > I would like to see some documentation about considerations about integrating tracing with other products (including the application server e.g. WildFly). What would be a good way to integrate span/parent with calls initated through casual inbound? Whats needs to be done when calls are started outside of casual (for instance through a HTTP inbound call or calls that originate from a domain that then need to perform casual outbound calls where you want tracing to work across from non-casual into the realm of casual? Integration is much easier if a vendor can point to best practices
> > > 
> > > 
> > > That's quite some undertaking perhaps, no idea so would need to spend quite some time to investigate and perhaps even experiment with what actually works - should we do that in this feature??? This feature is to be feature complete with casual 1.7 and protocol version up to 1.4.
> > > Maybe a POC to do what you suggest should be its own feature - if you do not already know that it is tiny, then that work can go into here as well?
> > 
> > 
> > I'm really not looking at anything super in depth here or a complete solution. Just quick pointers to stuff like (just giving examples, don't know if they would be considered correct/viable within the solution of tracing in casual-jca)
> > 
> > * Want to continue the trace outside of casual for an inbound call? Adopt the span/parent from `InboundContextScope::getContext`
> > * Got a call from outside of casual and want to continue a trace with casual? Create a context in `InboundContextScope` and add your context values to ensure outbound calls use your context, you are required to clear the context you create
> > * Span/parent id - what values to use for spanId, parentName and execution
> > 
> > But maybe I'm out of scope for this PR and that sort of work would be a separate task that looks deeper into the aforementioned
> 
> My feeling is that it is out of scope for this but that it should surely be done. Maybe you want to create an issue and perhaps even look at it at some point as I know you've already done quite a bit of work work with open telemetry and wildfly? That would then be complete with actual working examples how to integrate it.

Sure, I consider this resolved and created #199

cklellie (approved):

PR #202, opened 2026-04-07
feat: user defined code

aliothe:
Service call events should include user defined code when using gateway protocol version `>=` 1.3


casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/ServiceCallWorkListener.java
```diff
@@ -98,6 +98,11 @@ private ServiceCallEvent createEvent(Work work )
         {
             eventBuilder.withParentSpanId(message.getParentSpan().asHex())
                         .withSpanId(spanId.asHex());
+            if(work instanceof CasualServiceCallWork casualWork)
```
cklellie: Need to also check on TpNoReply as on line 107.

cklellie (approved):

PR #203, opened 2026-04-16
fix: Error handling for inbound

### LINKED ISSUE #101: CasualMessageListenerImpl  - service call request, handle WorkException
Currently the network connection would be closed.
The client would expect an error ( see ErrorState).

TobiasLeo:
Fixes several possible inbound-issues where exceptions are thrown:
- Service handler throws, make sure a correct error code is set ensuring that some error response is sent to casual.
- CasualServiceCallWork::issueCall throws outside of try, now JCA work listener handles thrown exception.

Both fixes add better error handling and ensure the callee casual domain does not hang waiting for a reply (at least until a timeout is reached).

Refs: #96 #101

### DIFF
--- a/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
+++ b/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
@@ -26,6 +26,7 @@ import se.laz.casual.network.protocol.messages.service.CasualServiceCallReplyMes
 import se.laz.casual.network.protocol.messages.service.CasualServiceCallRequestMessage;
 
 import java.util.UUID;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static se.laz.casual.network.ProtocolVersion.VERSION_1_3;
@@ -72,7 +73,7 @@ public final class CasualServiceCallWork implements Work
     @Override
     public void release()
     {
-        /**
+        /*
          * Currently no way to stop a service lookup or call.
          * Transaction context with which this Work is started
          * is applied with timeout that lies outside this code.
@@ -133,6 +134,13 @@ public final class CasualServiceCallWork implements Work
             replyBuilder.setError( ErrorState.TPENOENT )
                         .setTransactionState( TransactionState.ROLLBACK_ONLY );
             log.warning( ()-> "ServiceHandler not available for: " + message.getServiceName() );
+        }
+        catch( Throwable t)
+        {
+            replyBuilder.setError( ErrorState.TPESYSTEM )
+                        .setTransactionState( TransactionState.ROLLBACK_ONLY );
+            // This shouldn't happen with a well-behaved service handler. If it does, the handler that threw might need fixing.
+            log.log( Level.SEVERE, "An exception was thrown while handing a service call", t );
         }
         finally
         {
--- a/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
+++ b/casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
@@ -45,6 +45,7 @@ public final class CasualServiceCallWork implements Work
     private CasualNWMessage<CasualServiceCallReplyMessage> response;
     private ServiceHandler handler = null;
     private final SpanId spanId;
+    private boolean workFailedUnexpectedly = false;
 
     public CasualServiceCallWork(UUID correlationId, CasualServiceCallRequestMessage message, boolean isTpNoReply, ProtocolVersion protocolVersion, SpanId spanId)
     {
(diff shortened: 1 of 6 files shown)


casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/ServiceCallWorkListener.java
```diff
@@ -93,20 +137,24 @@ private ServiceCallEvent createEvent(CasualServiceCallWork work)
...
                         .withSpanId(spanId.asHex());
-            if(!isTpNoReply)
+            if(!isTpNoReply && workContainsReply(work))
             {
                 CasualServiceCallReplyMessage reply = work.getResponse().getMessage();
+
                 eventBuilder.withUserCode(reply.getUserDefinedCode());
             }
         }
         if(!isTpNoReply)
```
aliothe, line 152: Should not other branch also fail with TPESYSTEM in case there is an exception.
That is, if it is TPNOREPLY and we got an exception - I would expect TPESYSTEM as well.

casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
```diff
@@ -134,6 +135,13 @@ private void issueCall()
                         .setTransactionState( TransactionState.ROLLBACK_ONLY );
             log.warning( ()-> "ServiceHandler not available for: " + message.getServiceName() );
         }
+        catch( Throwable t)
```
cklellie, line 145: Do we also need a similar handling for no reply variant `issueCallNoReply`? In the case where there is a misbehaving service handler which handles a no reply call?

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/ServiceCallWorkListener.java
```diff
@@ -58,31 +69,64 @@ public ServiceCallWorkListener(Channel channel, CasualServiceCallRequestMessage
...
+        log.finest(() -> "Work started in %d ms with corrid %s".formatted(e.getStartDuration(), getCasualCorrelationId(getCasualServiceCallWork(e))));
         eventBuilder.start();
     }
 
     @Override
     public void workCompleted(WorkEvent e)
     {
         eventBuilder.end();
-        CasualServiceCallWork work = (CasualServiceCallWork) e.getWork();
+        if (e.getException() != null)
```
cklellie, line 114: Perhaps add a `log.finest` here too similar to in `workStarted`?

casual/casual-inbound-api/src/test/groovy/se/laz/casual/jca/inflow/work/CasualServiceCallWorkTest.groovy
```diff
@@ -253,4 +253,26 @@ class CasualServiceCallWorkTest extends Specification
         then:
         reply.getMessage().getError() == ErrorState.TPENOENT
     }
+
+    def "Call service where service handler throws an exception, returns result with TPESYSTEM status."()
```
cklellie, line 257: Maybe add a test with no reply variant ensuring noException thrown?

cklellie (commented):
Looks really good, just a few bits I have commented on if you could have a look at please.

aliothe (commented):
TPNOREPLY with exception, we would want TPESYSTEM in the event for that as well.

TobiasLeo:
> TPNOTRAN with exception, we would want TPESYSTEM in the event for that as well.

@aliothe I'm not sure I understand, where would TPNOTRAN not receive a TPESYSTEM if an unexpected error occurs as opposed to other scenarios?

casual/casual-inbound-api/src/main/java/se/laz/casual/jca/inflow/work/CasualServiceCallWork.java
TobiasLeo: I did this but could only log it as severe. A consequence was the need to introduce a new boolean in the work to flag it as unexpectedly failed to be able to handle the comment @aliothe had about reporting the error code in this event as TPESYSTEM and not OK

casual/casual-inbound/src/main/java/se/laz/casual/jca/inflow/ServiceCallWorkListener.java
TobiasLeo: Added this to ensure parity in logging between the different events

casual/casual-inbound-api/src/test/groovy/se/laz/casual/jca/inflow/work/CasualServiceCallWorkTest.groovy
TobiasLeo: Added, which also assumes we go with the solution I did now with catching exceptions and logging in TPNOREPLY scenario

cklellie (approved):

PR #204, opened 2026-04-21
chore: networking cleanup

cklellie:
Cleanup the networking code to remove unused code and reuse common elements.
Bump sonar to latest version and resolve outstanding issues.


casual/casual-api/src/main/java/se/laz/casual/api/xa/XIDFormatType.java
aliothe, line 1: What are the actual changes here?

casual/casual-api/src/main/java/se/laz/casual/network/ProtocolVersionException.java
```diff
@@ -10,6 +10,8 @@
 
 import java.util.function.Supplier;
 
+// java:S110 - Deep inheritance ok
+@SuppressWarnings( "java:S110" )
```
aliothe, line 14: Is this an old issue?

aliothe (approved):
Looks good

casual/casual-api/src/main/java/se/laz/casual/api/xa/XIDFormatType.java
cklellie: It was a sonar issue - "merge dandling javadoc" or something. Basically because it was two javadoc block one after the other rather than one, so i merged them with a break between.

casual/casual-api/src/main/java/se/laz/casual/network/ProtocolVersionException.java
cklellie: It was not new with this change, no.

PR #209, opened 2026-05-22
feat: Added singleton for jca info

Alex-Erik:
Added singleton which purpouse is to store information about casual jca for easier access. Its main consumer will be casual-java cli.

### DIFF
--- a/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
@@ -1,10 +1,11 @@
 package se.laz.casual.info;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Static instance that contains information about inbound Casual Services.
@@ -12,28 +13,25 @@ import java.util.Optional;
  */
 public final class CasualInfo
 {
-    private static CasualInfo instance;
-    private static final Map<String, Service> inboundServices = new HashMap<>();
+    private static final CasualInfo instance = new CasualInfo();
+    private final Map<String, Service> inboundServices;
 
-    private CasualInfo() {}
-
-    public static CasualInfo getInstance() {
-        if(instance == null) {
-            instance = new CasualInfo();
-        }
-        return instance;
+    private CasualInfo() {
+        this.inboundServices = new ConcurrentHashMap<>();
     }
 
-    public void addInboundService(Service service) {
-        inboundServices.put(service.getName(), service);
+    public static void addInboundService(Service service) {
+        Objects.requireNonNull(service, "service must not be null");
+        instance.inboundServices.put(service.getName(), service);
     }
 
-    public List<Service> getServices() {
-        return new ArrayList<>(inboundServices.values());
+    public static List<Service> getServices() {
+        return new ArrayList<>(instance.inboundServices.values());
     }
 
-    public Optional<Service> getInboundService(String name)
+    public static Optional<Service> getInboundService(String name)
     {
-        return Optional.ofNullable(inboundServices.get(name));
+        Objects.requireNonNull(name, "name must not be null");
+        return Optional.ofNullable(instance.inboundServices.get(name));
     }
 }
--- a/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
@@ -8,7 +8,8 @@ import java.util.Optional;
 import java.util.logging.Logger;
 
 /**
- * Static instance that contains information about inbound Casual Services.
+ * Static instance that stores information about Casual jca, services and queues
+ * in {@link CasualInfoStorage}.
  * Used primarily by casual-java cli.
  */
 public final class CasualInfo
@@ -26,17 +27,37 @@ public final class CasualInfo
         return instance;
     }
 
+    /**
+     * Add an inbound service to storage
+     *
+     * @param service - a service
+     */
     public void addService( Service service )
     {
         Objects.requireNonNull( service, "service must not be null" );
         CasualInfoStorage.getInstance().putService( service );
     }
 
+    /**
+     * Add an outbound service to storage
+     *
+     * @param serviceDetailsList - details about the service
+     * @param connection         - service connection details
+     */
     public void addDiscovery( List<ServiceDetails> serviceDetailsList, Connection connection )
     {
         serviceDetailsList.forEach( serviceDetails -> addDiscoveredService( serviceDetails, connection ) );
     }
 
+    /**
+     * Store and/or update statistics for a service.
+     * Should only be called when an event is created from a service call.
+     *
+     * @param serviceName - service name
+     * @param order       - order of service
+     * @param start       - start time
+     * @param end         - end time
+     */
     public void storeEvent( String serviceName, char order, long start, long end )
     {
         EventServiceStatistics eventServiceStatistics = toEvent( serviceName, order, start, end );
@@ -45,6 +66,12 @@ public final class CasualInfo
         LOG.finest( () -> "Stored statistics: '%s'.".formatted( eventServiceStatistics ) );
     }
 
+    /**
+     * Adds an outbound service to storage
+     *
+     * @param serviceDetails - service details
+     * @param connection     - service connection details
+     */
     private void addDiscoveredService( ServiceDetails serviceDetails, Connection connection )
     {
         ServiceDescriptor serviceDescriptor = new ServiceDescriptor( serviceDetails.getName(), Order.CONCURRENT );
@@ -55,6 +82,15 @@ public final class CasualInfo
         CasualInfoStorage.getInstance().putService( builder.build() );
     }
 
+    /**
+     * Creates or updates statistics for a specific service (inbound or outbound).
+     *
+     * @param serviceName - service
+     * @param order       - order of service e.g. inbound or outbound
+     * @param start       - start time
+     * @param end         -  end time
+     * @return EventServiceStatistics
+     */
     private EventServiceStatistics toEvent( String serviceName, char order, long start, long end )
     {
         EventServiceStatistics eventServiceStatistics =
--- a/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
@@ -1,0 +1,39 @@
+package se.laz.casual.info;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Static instance that contains information about inbound Casual Services.
+ * Used primarily by casual-java cli.
+ */
+public final class CasualInfo
+{
+    private static CasualInfo instance;
+    private static final Map<String, Service> inboundServices = new HashMap<>();
+
+    private CasualInfo() {}
+
+    public static CasualInfo getInstance() {
+        if(instance == null) {
+            instance = new CasualInfo();
+        }
+        return instance;
+    }
+
+    public void addInboundService(Service service) {
+        inboundServices.put(service.getName(), service);
+    }
+
+    public List<Service> getServices() {
+        return new ArrayList<>(inboundServices.values());
+    }
+
+    public Optional<Service> getInboundService(String name)
+    {
+        return Optional.ofNullable(inboundServices.get(name));
+    }
+}
--- a/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
@@ -1,16 +1,10 @@
 package se.laz.casual.info;
 
-import se.laz.casual.api.discovery.DiscoveryReturn;
-import se.laz.casual.api.queue.QueueDetails;
 import se.laz.casual.api.service.ServiceDetails;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Logger;
 
 /**
@@ -19,123 +13,61 @@ import java.util.logging.Logger;
  */
 public final class CasualInfo
 {
-    private static Logger LOG = Logger.getLogger( CasualInfo.class.getName() );
-
+    private static final Logger LOG = Logger.getLogger( CasualInfo.class.getName() );
     private static final CasualInfo instance = new CasualInfo();
-    private final Map<String, Service> inboundServices;
-    private final Map<String, List<Service>> outboundServices;
-    private final Map<String, EventServiceStatistics> inboundStatistics;
-    private final Map<String, EventServiceStatistics> outboundStatistics;
 
     private CasualInfo()
     {
-        this.inboundServices = new ConcurrentHashMap<>();
-        this.outboundServices = new ConcurrentHashMap<>();
-        this.inboundStatistics = new ConcurrentHashMap<>();
-        this.outboundStatistics = new ConcurrentHashMap<>();
+        // no-op
     }
 
-    public static void addInboundService( Service service )
+    public static CasualInfo getInstance()
+    {
+        return instance;
+    }
+
+    public void addService( Service service )
     {
         Objects.requireNonNull( service, "service must not be null" );
-        instance.inboundServices.put( service.getName(), service );
+        CasualInfoStorage.getInstance().putService( service );
     }
 
-    public static List<Service> getInboundServices()
+    public void addDiscovery( List<ServiceDetails> serviceDetailsList, Connection connection )
     {
-        return instance.inboundServices.values().stream().toList();
+        serviceDetailsList.forEach( serviceDetails -> addDiscoveredService( serviceDetails, connection ) );
     }
 
-    public static Optional<Service> getInboundService( String serviceName )
+    public void storeEvent( String serviceName, char order, long start, long end )
     {
-        Objects.requireNonNull( serviceName, "name must not be null" );
-        return Optional.ofNullable( instance.inboundServices.get( serviceName ) );
-    }
-
-    public static Optional<EventServiceStatistics> getInboundStatistic( String serviceName )
-    {
-        return Optional.ofNullable( instance.inboundStatistics.get( serviceName ) );
-    }
-
-    public static List<Service> getOutboundServices()
-    {
-        List<Service> outboundServices = new ArrayList<>();
-        instance.outboundServices.values().forEach( outboundServices::addAll );
-        return outboundServices;
-    }
-
-    public static List<Service> getOutboundService( String serviceName )
-    {
-        Objects.requireNonNull( serviceName, "name must not be null" );
-        return instance.outboundServices.getOrDefault( serviceName, Collections.emptyList() );
-    }
-
-    public static Optional<EventServiceStatistics> getOutboundStatistic( String serviceName )
-    {
-        return Optional.ofNullable( instance.outboundStatistics.get( serviceName ) );
-    }
-
-    public static Map<String, EventServiceStatistics> getInboundStatistics()
-    {
-        return instance.inboundStatistics;
-    }
-
-    public static Map<String, EventServiceStatistics> getOutboundStatistics()
-    {
-        return instance.outboundStatistics;
-    }
-
-    public static void addDiscovery( DiscoveryReturn discoveryReturn, Connection connection )
-    {
-        discoveryReturn.getQueueDetails().forEach( queueDetails -> addOutboundQueue( queueDetails, connection ) );
-        discoveryReturn.getServiceDetails().forEach( serviceDetails -> addOutboundService( serviceDetails,
-                connection ) );
-    }
-
-    public static void addDiscovery( List<ServiceDetails> serviceDetailsList, Connection connection )
-    {
-        serviceDetailsList.forEach( serviceDetails -> addOutboundService( serviceDetails, connection ) );
-    }
-
-    public static void storeEvent( String serviceName, char order, long start, long end )
-    {
-        EventServiceStatistics eventServiceStatistics = instance.inboundStatistics.getOrDefault( serviceName,
-                new EventServiceStatistics.Builder().name( serviceName )
-                        .order( order ).build() );
-        eventServiceStatistics.increment();
-        long executionTime = end - start;
-        eventServiceStatistics.setMin( executionTime );
-        eventServiceStatistics.setMax( executionTime );
-        eventServiceStatistics.setLast( end );
-        eventServiceStatistics.increaseTotal( executionTime );
-        if( order == 'S' )
-        {
-            instance.inboundStatistics.put( serviceName, eventServiceStatistics );
-        }
-        else
-        {
-            instance.outboundStatistics.put( serviceName, eventServiceStatistics );
-        }
+        EventServiceStatistics eventServiceStatistics = toEvent( serviceName, order, start, end );
+        CasualInfoStorage.getInstance().putEvent( new ServiceDescriptor( eventServiceStatistics.getName(),
+                Order.unmarshall( order ) ), eventServiceStatistics );
         LOG.finest( () -> "Stored statistics: '%s'.".formatted( eventServiceStatistics ) );
     }
 
-    private static void addOutboundService( ServiceDetails serviceDetails, Connection connection )
+    private void addDiscoveredService( ServiceDetails serviceDetails, Connection connection )
     {
-        Service service = new Service.Builder().newBuilder( serviceDetails )
-                .connection( connection ).build();
-        String domainId = connection.getDomainId().getId().toString();
-        List<Service> allDiscoveries = instance.outboundServices.getOrDefault( domainId,
-                Collections.synchronizedList( new ArrayList<>() ) );
-        Optional<Service> exists = allDiscoveries.stream().findFirst().filter( s -> s.equals( service ) );
-        if( exists.isEmpty() )
-        {
-            allDiscoveries.add( service );
-            instance.outboundServices.put( domainId, allDiscoveries );
-        }
+        ServiceDescriptor serviceDescriptor = new ServiceDescriptor( serviceDetails.getName(), Order.CONCURRENT );
+        Optional<Service> service = CasualInfoStorage.getInstance().getService( serviceDescriptor );
+        Service.Builder builder = service.isPresent() ? Service.newBuilder( service.get() ) :
+                Service.newBuilder( serviceDetails, Order.CONCURRENT );
+        builder.connection( connection );
+        CasualInfoStorage.getInstance().putService( builder.build() );
     }
 
-    private static void addOutboundQueue( QueueDetails queueDetails, Connection connection )
+    private EventServiceStatistics toEvent( String serviceName, char order, long start, long end )
     {
-        //Not implemented!
+        EventServiceStatistics eventServiceStatistics =
+                CasualInfoStorage.getInstance().getServiceStatistic( new ServiceDescriptor( serviceName,
+                                Order.unmarshall( order ) ) )
+                        .orElseGet( () -> new EventServiceStatistics.Builder().name( serviceName ).order( order ).build() );
+        EventServiceStatistics.Builder builder = EventServiceStatistics.newBuilder( eventServiceStatistics );
+        long executionTime = end - start;
+        builder.increment();
+        builder.max( executionTime );
+        builder.min( executionTime );
+        builder.last( start );
+        builder.increaseTotal( executionTime );
+        return builder.build();
     }
 }
--- a/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
+++ b/casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
@@ -1,11 +1,17 @@
 package se.laz.casual.info;
 
+import se.laz.casual.api.discovery.DiscoveryReturn;
+import se.laz.casual.api.queue.QueueDetails;
+import se.laz.casual.api.service.ServiceDetails;
+
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Logger;
 
 /**
  * Static instance that contains information about inbound Casual Services.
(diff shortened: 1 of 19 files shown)


casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Optional;
+
+/**
+ * Static instance that contains information about inbound Casual Services.
+ * Used primarily by casual-java cli.
+ */
+public final class CasualInfo
+{
+    private static CasualInfo instance;
+    private static final Map<String, Service> inboundServices = new HashMap<>();
```
cklellie: Use `ConcurrentHashMap` to prevent issues with updates from multiple threads.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private CasualInfo() {}
+
+    public static CasualInfo getInstance() {
+        if(instance == null) {
+            instance = new CasualInfo();
```
cklellie: Better practise for thread safety to declare new instance inline with the static field declaration rather than using lazy initilisation.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,164 @@
+package se.laz.casual.info;
+
+import se.laz.casual.network.messages.domain.TransactionType;
+
+import java.util.Objects;
+
+public class Service
```
cklellie, line 11: This will be our fourth Service "data" object. `ServiceDetails`, `ServiceInfo`, `Service` and now this one also called `Service`. Not ideal. Though I understand you want to keep it seperate right now.
Was the idea this object would also later include the metrics aggregation data too or will that be seperate?

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,164 @@
...
+    {
+        return registred;
+    }
+
+    public String getJndiName()
+    {
+        return jndiName;
+    }
+
+    public void setRegistred(boolean registred)
```
cklellie: Immutable data types are preferable. Particularly when you are passing these objects out to those outside your control. E.g. getServices().
So a builder with new from original source is preferable. e.g `Service.newBuilder( src )....build()`

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,164 @@
...
+                ", category='" + category + '\'' +
+                ", hops=" + hops +
+                ", registred=" + registred +
+                ", jndiName='" + jndiName + '\'' +
+                ", transactionType=" + transactionType +
+                ", timeout=" + timeout +
+                '}';
+    }
+
+    public static class Builder {
```
cklellie: The style of builder we have been using so far in the code base, we provide a static method for creating a new instance e.g. `Service.newBuilder()....build()`.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,164 @@
...
+        this.timeout = builder.timeout;
+        this.transactionType = builder.transactionType;
+        this.hops = builder.hops;
+        this.registred = builder.registred;
+        this.jndiName = builder.jndiName;
+
+        Objects.requireNonNull( name );
+        Objects.requireNonNull( category );
+        Objects.requireNonNull( transactionType );
+        Objects.requireNonNull( jndiName );
```
cklellie, line 39: What about external services - which potentially don't have jndiName?

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+        return instance;
+    }
+
+    public void addInboundService(Service service) {
+        inboundServices.put(service.getName(), service);
```
cklellie: null check param service.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public List<Service> getServices() {
+        return new ArrayList<>(inboundServices.values());
+    }
+
+    public Optional<Service> getInboundService(String name)
+    {
+        return Optional.ofNullable(inboundServices.get(name));
```
cklellie: null check param name.

cklellie (changes requested):
Nice size PR. Few issues commented. 
Though main question is around the future development of this "service data" object you have, hopefully over time you will see the usage patterns you need and we will be able to better determine what should be included before we merge the `feature/cli/dev` branch into `dev`.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
Alex-Erik: Yes. All of these service-objects should probably be merged into one. 

The metric aggregation will be done in the casual-java-cli-app, not in jca itself, and stored in a separate data structure.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
Alex-Erik: Casual-jca do not know of external services, if I am not mistaken. It is casual-caller that discovers external services and stores information about them. So all services here should have a jndi-name.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -1,39 +1,37 @@
...
-    public void addInboundService(Service service) {
-        inboundServices.put(service.getName(), service);
+    public static void addInboundService(Service service) {
+        Objects.requireNonNull(service, "service must not be null");
+        instance.inboundServices.put(service.getName(), service);
     }
 
-    public List<Service> getServices() {
-        return new ArrayList<>(inboundServices.values());
+    public static List<Service> getServices() {
```
cklellie: Leave these as member methods, as it improves testability later. For example, you can't easily mock static in spock, which will likely be useful for testing from other places later.

Alex-Erik:
With the latest changes on casual-jca and CasualInfo, the following is shown for `casual-java services -i`:

Screenshot_20260610_152138

And `casual-java services -ls`:
Screenshot_20260610_152538

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+        return outboundServices;
+    }
+
+    public static List<Service> getOutboundService( String serviceName )
+    {
+        Objects.requireNonNull( serviceName, "name must not be null" );
+        return instance.outboundServices.getOrDefault( serviceName, Collections.emptyList() );
+    }
+
+    public static Optional<EventServiceStatistics> getOutboundStatistic( String serviceName )
```
cklellie: Why the seperation of inbound and outbound?

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+                        .order( order ).build() );
+        eventServiceStatistics.increment();
+        long executionTime = end - start;
+        eventServiceStatistics.setMin( executionTime );
+        eventServiceStatistics.setMax( executionTime );
+        eventServiceStatistics.setLast( end );
+        eventServiceStatistics.increaseTotal( executionTime );
+        if( order == 'S' )
+        {
+            instance.inboundStatistics.put( serviceName, eventServiceStatistics );
```
cklellie: Update race condition not handled here. Potential for loss of info.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public static Optional<EventServiceStatistics> getOutboundStatistic( String serviceName )
+    {
+        return Optional.ofNullable( instance.outboundStatistics.get( serviceName ) );
+    }
+
+    public static Map<String, EventServiceStatistics> getInboundStatistics()
+    {
+        return instance.inboundStatistics;
```
cklellie: Defensive copy returns, otherwise caller can modify the internal state.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public static Map<String, EventServiceStatistics> getOutboundStatistics()
+    {
+        return instance.outboundStatistics;
```
cklellie: Defensive copy returns.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
cklellie: Additional EventServiceStatistic being mutable is problematic as it is returned here.
Immutability is more often the simplest way to deal with maintaining updating state.

casual/casual-api/src/main/java/se/laz/casual/info/EventServiceStatistics.java
```diff
@@ -0,0 +1,151 @@
+package se.laz.casual.info;
+
+import java.util.Objects;
+
+public class EventServiceStatistics
+{
+    String name;
+    char order;
+    long count;
+    long total;
```
cklellie: This class should probably be immutable with updates creating new instances.
Alternatively, count and total should be AtomicLong or AtomicInteger and synchronisation around min max last updates.
Though you will still have the issue of how you do defensive returns if you leave this object mutable.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,195 @@
...
+
+import java.util.Objects;
+
+public class Service
+{
+    private final String name;
+    private final String category;
+    private final long hops;
+    private boolean registered;
+    private final String jndiName;
```
cklellie, line 18: jndiname is still mandatory, why?
Also after "resolved" jndi is no longer set, so how does this work?
If you start the server, check the services, deploy a new app, i would want the services in the new app to be visible and with the correct jndiName is applicable.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
cklellie: It is both.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,195 @@
...
+            return this;
+        }
+
+        public Builder connection( Connection connection )
+        {
+            this.connection = connection;
+            return this;
+        }
+
+        public Builder newBuilder( Service src )
```
cklellie: Pattern we use for `newBuilder` - it should be static on the outer class not a member method on the inner class. The builder should also have private Constructor to prevent invocation incorrectly.
So you have `Outclass.newBuilder().build()`, not `Outclass.Builder().build()`.
Or `Outclass.newBuilder( src ).build()` for copy variant.
So they are consistent.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
```diff
@@ -43,6 +44,7 @@ public void post(ServiceCallEvent event)
         try
         {
             handler.put(event);
+            CasualInfo.storeEvent( event.getService(), event.getOrder(), event.getStart(), event.getEnd() );
```
cklellie: Update for testing? Quick check with sonar shows 8.8% coverage on new code.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.Logger;
+
+/**
+ * Static instance that contains information about inbound Casual Services.
+ * Used primarily by casual-java cli.
+ */
+public final class CasualInfo
+{
+    private static Logger LOG = Logger.getLogger( CasualInfo.class.getName() );
```
cklellie: Add final - `private static final Logger`

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+        this.outboundStatistics = new ConcurrentHashMap<>();
+    }
+
+    public static void addInboundService( Service service )
+    {
+        Objects.requireNonNull( service, "service must not be null" );
+        instance.inboundServices.put( service.getName(), service );
+    }
+
+    public static List<Service> getInboundServices()
```
cklellie: Service is mutable, so even though you are returning a copy of the list, it's contents is still mutable.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private CasualInfo()
+    {
+        this.inboundServices = new ConcurrentHashMap<>();
+        this.outboundServices = new ConcurrentHashMap<>();
+        this.inboundStatistics = new ConcurrentHashMap<>();
+        this.outboundStatistics = new ConcurrentHashMap<>();
+    }
+
+    public static void addInboundService( Service service )
```
cklellie: Why the difference for inbound and outbound add?

Inbound is addInboundService, whilst outbound is a private addOutboundService with different transfer option from addDiscovery.

This class is at risk of doing too much. It would perhaps be nicer to break down into a smaller set of functionality.
This should be the end destination for things, transformed outside (with helper methods) rather than inside to make the seperation of concerns cleaner:

service, queue, config

* `addXXX`
* `getXXX( String name )`
* `getAllXXX`

Then additional method of metrics storage, rather than events. As you have correctly sent the decomposed object in, not the original event object.
Currently it is just service we have, but I can see queue also being a useful metric to have.

* `addXXXMetric`
* `getXXXMetrics( String name )`
* `getAllXXXMetric`

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,141 @@
...
+        discoveryReturn.getServiceDetails().forEach( serviceDetails -> addOutboundService( serviceDetails,
+                connection ) );
+    }
+
+    public static void addDiscovery( List<ServiceDetails> serviceDetailsList, Connection connection )
+    {
+        serviceDetailsList.forEach( serviceDetails -> addOutboundService( serviceDetails, connection ) );
+    }
+
+    public static void storeEvent( String serviceName, char order, long start, long end )
```
cklellie: What about pending?

casual/casual-jca/src/main/java/se/laz/casual/jca/service/CasualServiceCaller.java
```diff
@@ -209,6 +211,12 @@ public List<ServiceDetails> serviceDetails(String serviceName)
                                 .withTimeout(service.getTimeout())
                                 .withHops(service.getHops()).build()));
 
+        CasualInfo.addDiscovery(serviceDetailsList, new Connection.Builder()
```
cklellie: This should changed to call the CasualDiscoveryCaller - rather than duplicate the storage of the discovery.

casual/casual-jca/src/main/java/se/laz/casual/jca/CasualManagedConnection.java
```diff
@@ -108,6 +108,11 @@ private NetworkConnection getOrCreateFromPool()
                                          mcf.getNetworkConnectionPoolSize());
     }
 
+    public CasualManagedConnectionFactory getMcf()
```
cklellie: naming - `getMangagedConnectionFactory`

cklellie (changes requested):
Please look over multi threading considerations for metrics and prevention of leaking out mutable data, preferably with immutability.
Also there is very little unit testing or javadoc documentation.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
Alex-Erik: I will add more tests. Wanted to push the latest changes so I could get some feedback first.

casual/casual-jca/src/main/java/se/laz/casual/jca/service/CasualServiceCaller.java
Alex-Erik: Yes. However, this code may be used by others and therefore we must still push data to CasualInfo as long as this method exists.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
Alex-Erik: I differentiated inbound and outbound because how else can we know which is which? An inbound service can have the same name as an outbound service and I have not found anything that can be used as a "identifier". 

I can rearrange the code into smaller methods.

Do you want me to rename everything to metric instead of event?

Yes, queue metrics will be added but this pr is about list services.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
Alex-Erik: jndiName is en empty string by default. Do you want it to be null instead?

Not really sure what you mean with that jndi is no longer set? I realize now that it should be set earlier in CasualServiceRegistry, but it wont be unset as it is now. Can the jndiName change between CasualServiceRegistry and JndiSerachTimer?

casual/casual-jca/src/main/java/se/laz/casual/jca/service/CasualServiceCaller.java
cklellie: Correct, but it will do that still by calling CasualDiscoveryCaller - as you have added the push of data to CasualInfo now there too.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
cklellie: Yes, the purpose of JndiSearchTimer is if the jndi name is not determined at bean discovery, it will search for it to find the true jndi name.
I found the part of JndiSearchTimer line 93-98 now:
```java
                Optional<Service> inboundService = CasualInfo.getInboundService(found.getServiceName());
                inboundService.ifPresent( service -> CasualInfo.addInboundService(
                        new Service.Builder().newBuilder( service )
                                .registred( true )
                                .jndiName( found.getJndiName() )
                                .build()
```
So, with this you are using a copy builder, so there is no need to call set on registered. Checking in intellij, it finds no usage of setRegistered. Which is what was confusing me, as you only removed set jndi. So if you remove setter for registered and then registered can be final and this class can be fully immutable.

jndi being empty or null per default - it depends really. Outbound will never have this value, so perhaps it should be null to allow optional return for that? Though equally as with registered that is only valid on inbound too, so outbound is defaulting to registered false, which is potentially confusing. Registered is effectively a computed value determined by whether the service is inbound and has a value for jndi other than null. Otherwise it is optional empty. I don't know exactly the right approach here, but if you make everything immutable first and we can think about the other part later.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
cklellie: We know if they are inbound or outbound depending on how we find them. If via discovery - outbound, if via bean discovery - inbound.

Equally, I see you are using the domainId on the connection for outbound, you could also retrieve the current local domain from configuration for the inbound. That would also differenciate them.

I realise now that the key in the outbound services map is not the service name, but the domain id for outbound, which is confusing, as you have the method `getOutboundService( String serviceName )` which uses service name as the map key to return - so that won't work currently and will just return an empty collection. I see that it is not currently used, so that explains why perhaps.

How we store the data needs to be determine by the usage patterns we see for the data. I see the keys in maps being the index for what we want to retrieve based on some query criteria to aid efficient retrieval or modification.

I am unsure when we will need to differenciate between inbound and outbound, in terms of retrieving just inbound or just outbound services? If there is none, then it feels like it should just be additional meta data about the services.
There is the "order" concept, which is used currently in events. Perhaps order as a field on the Service type - which has a new enum value of either Inbound or Outbound?

Equally, I don't know where we need to know which domain something came from in terms of returning only domain x services? Is that a functionality in the cli? 

Domain id might be useful with regard to instance count though? If you have multiple domains with the same service name. I have not tested what happens if you have a service with same name both inbound and outbound when listed via casual cli, that would be good to check.

Again if not needed for retrieval or update purposes - is domainId  just meta data on the Service object? A list of domainIds? Or should it be a single Service entry one for each domainId? Or even just a count ?No, probably not a count. Though that is an interesting question about how we know when a service is no longer available at a specific domain - i think caller push might be required there. Though it might be possible also within discovery - to diff the results of the discovery, if you asked about a named service but it didn't find it - that would show it is not there - which if we had a record of it - would show it was now gone. That would be based on domainId - so perhaps there is a need for domainId being more promenant.

I think a bit more thought around the data model will be beneficial in the long run, as you can see from my ramblings - there are quite a few open questions still.

Metrics vs Events - not sure - I can argue both ways. Leave as event for now.

casual/casual-event-api/src/main/java/se/laz/casual/event/ServiceCallEventPublisher.java
cklellie: Ok, but as a TDD type person - I love to read unit tests to see how the code is being/ going to be used in lue of detailed documentation, otherwise I often have many more questions about why things are the way they are :)

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
Alex-Erik: I have removed the separation

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfoStorage.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Static instance that contains information about inbound Casual Jca, services and queues.
+ * Used primarily by casual-java cli.
```
cklellie, line 20: I like the seperation here, but I think CasualInfo should be the entry point. I see you have protected methods here, which I guess is because you don't want to allow others to access those methods.
If you shield this behind the CasualInfo as a facade, that would provide the same functionality. Could even make this class default package access only too.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,109 @@
...
+ * Used primarily by casual-java cli.
+ */
+public final class CasualInfo
+{
+    private static final Logger LOG = Logger.getLogger( CasualInfo.class.getName() );
+    private static final CasualInfo instance = new CasualInfo();
+
+    private CasualInfo()
+    {
+        // no-op
```
cklellie, line 21: You could simplify the store (CasualInfoStorage) access, by initialising it from the singleton once here and using that field throughout.

casual/casual-api/src/main/java/se/laz/casual/info/Order.java
```diff
@@ -0,0 +1,44 @@
...
+package se.laz.casual.info;
+
+import java.util.Arrays;
+
+//TODO: Should merge with event-api Order
+public enum Order
+{
+    // note: really means inbound
+    // It is the same way as it is in casual
+    SEQUENTIAL( 'S' ),
```
cklellie, line 15: Yes, this namiing is confusing - perhaps we can use INBOUND and OUTBOUND here instead?

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfoStorage.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Add/update service in storage
+     *
+     * @param service - service to be added/updated
+     */
+    protected void putService( Service service )
+    {
+        Objects.requireNonNull( service, "service must not be null" );
+        services.put( new ServiceDescriptor( service.getName(), service.getOrder() ),
+                Collections.synchronizedList( List.of( service ) ) );
```
cklellie, line 86: This is overwriting, not accumulating per connection? 
Also in the getService - you are only returning the index 0 value?

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,109 @@
...
+        CasualInfoStorage.getInstance().putService( service );
+    }
+
+    /**
+     * Add an outbound service to storage
+     *
+     * @param serviceDetailsList - details about the service
+     * @param connection         - service connection details
+     */
+    public void addDiscovery( List<ServiceDetails> serviceDetailsList, Connection connection )
```
cklellie, line 46: Why do we need two ways to addService?
Inbound don't have a connection per se - but that is just meta on the Service now.
I don't see why we need the addition retrieval check in addDiscoveredService()?

Perhaps each service is unique based on name and connection effectively? Be that local (inbound) or external host:port (outbound).

Then metrics should be associated with the service:connection pair - with `casual service -ls` showing aggregation of all, whilst `casual service --list-instances` currently doesn't show metrics per instance, but could.

The connection is not part of the event, but it is available near to where they are created - so that could be rectified easily.

casual/casual-api/src/main/java/se/laz/casual/info/CasualInfo.java
```diff
@@ -0,0 +1,109 @@
...
+     *
+     * @param serviceName - service name
+     * @param order       - order of service
+     * @param start       - start time
+     * @param end         - end time
+     */
+    public void storeEvent( String serviceName, char order, long start, long end )
+    {
+        EventServiceStatistics eventServiceStatistics = toEvent( serviceName, order, start, end );
+        CasualInfoStorage.getInstance().putEvent( new ServiceDescriptor( eventServiceStatistics.getName(),
```
cklellie: This is still not thread safe.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,213 @@
...
+
+    public static Builder newBuilder( Service src )
+    {
+        return new Builder()
+                .name( src.getName() )
+                .order( src.getOrder() )
+                .category( src.getCategory() )
+                .hops( src.getHops() )
+                .registred( src.isRegistered() )
+                .jndiName( src.getJndiName() );
```
cklellie, line 95: Missing connection.

casual/casual-api/src/main/java/se/laz/casual/info/Service.java
```diff
@@ -0,0 +1,213 @@
...
+
+    public static Builder newBuilder( ServiceDetails src, Order order )
+    {
+        return new Builder()
+                .name( src.getName() )
+                .order( order )
+                .transactionType( src.getTransactionType() )
+                .timeout( src.getTimeout() )
+                .category( src.getCategory() )
+                .hops( src.getHops() );
```
cklellie, line 106: missing connection.

cklellie (commented):
The modelling for services is better, but still feels like it is not quite there for the different use cases that we will have going forward.
I think connection is more prominent than inbound and outbound, for example, there can be many different outbound connections with the same exact service name. This is aggregated in the `casual service --list-services` view, but visible with the `casual service --list-instances` view.
How metrics are then stored is likely to be related to the connections, with the cli summarising the view over all connections I would imagine.

cklellie:
Example output where inbound and outbound have same service names.
```shell
/test/casual $ casual service --list-services
name                                                         category  V  mode    timeout  contract  I  C  AT     min    max    P  PAT    RI  RC  last                            
-----------------------------------------------------------  --------  -  ------  -------  --------  -  -  -----  -----  -----  -  -----  --  --  --------------------------------
casual/example/advertised/echo                               example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/atomic/echo                                   example   D  atomic        -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/auto/echo                                     example   D  auto          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/conversation                                  example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/conversation_recv_send                        example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/conversation_recv_send_auto                   example   D  auto          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/domain/echo/97de73bb28e8476da0c35ee4eb68a2a3  example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/domain/echo/test-domain                       example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/domain/name                                   example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/echo                                          example   D  join          -         -  1  2  0.000  0.000  0.000  0  0.000   1   0  2026-06-16T11:34:16.183813+00:00
casual/example/error/system                                  example   D  join          -         -  1  1  0.000  0.000  0.000  0  0.000   0   0  2026-06-16T11:34:16.473866+00:00
casual/example/forward                                       example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/forward/echo                                  example   D  auto          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/forward/join/echo                             example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   1   0  -                               
casual/example/lowercase                                     example   D  join          -         -  1  1  0.000  0.000  0.000  0  0.000   0   0  2026-06-16T11:34:16.333036+00:00
casual/example/rollback                                      example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/sink                                          example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/sleep                                         example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/terminate                                     example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/undiscoverable/echo                           example   U  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
casual/example/uppercase                                     example   D  join          -         -  1  1  0.000  0.000  0.000  0  0.000   0   0  2026-06-16T11:34:16.270100+00:00
casual/example/work                                          example   D  join          -         -  1  0  0.000  0.000  0.000  0  0.000   0   0  -                               
/test/casual $ casual service --list-instances
service                                                      state     hops  pid  alias                  description
-----------------------------------------------------------  --------  ----  ---  ---------------------  -----------
casual/example/advertised/echo                               idle         0   12  casual-example-server  -          
casual/example/atomic/echo                                   idle         0   12  casual-example-server  -          
casual/example/auto/echo                                     idle         0   12  casual-example-server  -          
casual/example/conversation                                  idle         0   12  casual-example-server  -          
casual/example/conversation_recv_send                        idle         0   12  casual-example-server  -          
casual/example/conversation_recv_send_auto                   idle         0   12  casual-example-server  -          
casual/example/domain/echo/97de73bb28e8476da0c35ee4eb68a2a3  idle         0   12  casual-example-server  -          
casual/example/domain/echo/test-domain                       idle         0   12  casual-example-server  -          
casual/example/domain/name                                   idle         0   12  casual-example-server  -          
casual/example/echo                                          idle         0   12  casual-example-server  -          
casual/example/echo                                          external     0   14  casual-outbound        test-domain
casual/example/error/system                                  idle         0   12  casual-example-server  -          
casual/example/forward                                       idle         0   12  casual-example-server  -          
casual/example/forward/echo                                  idle         0   12  casual-example-server  -          
casual/example/forward/join/echo                             idle         0   12  casual-example-server  -          
casual/example/forward/join/echo                             external     0   14  casual-outbound        test-domain
casual/example/lowercase                                     idle         0   12  casual-example-server  -          
casual/example/rollback                                      idle         0   12  casual-example-server  -          
casual/example/sink                                          idle         0   12  casual-example-server  -          
casual/example/sleep                                         idle         0   12  casual-example-server  -          
casual/example/terminate                                     idle         0   12  casual-example-server  -          
casual/example/undiscoverable/echo                           idle         0   12  casual-example-server  -          
casual/example/uppercase                                     idle         0   12  casual-example-server  -          
casual/example/work                                          idle         0   12  casual-example-server  -
```

PR #211, opened 2026-06-15
feat: changes for quarkus

aliothe:
Added an API to add handlers found via SPI - this is needed for the quarkus extension.

The quarkus extension still uses SPI to find any handlers but in dev mode it also needs to register them.
It is only used by the quarkus extension, in a Java application server - everything works as before.

Expose the inbound transaction registry via CasualResourceAdapter - used by the quarkus extension.

Changes to robustness found during soak testing:
* expose an API to know if the connection is currently disconnecting - used by casual caller to not issue any service/queue calls
* use the pool address instead of domain id in XAResource start/end - there is no need for any network call here and it serves the same purpose
* in the NettyNetworkConnection - when the domain is disconnecting, only messages allowed are XA messages.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,9 +43,9 @@ allprojects{
           content {
               includeGroup("se.laz.casual")
           }
-      }
+        }
+      mavenLocal()
       mavenCentral()
-
   } 
 }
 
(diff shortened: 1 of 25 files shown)


cklellie (approved):
Looks good to me.

